PR #9, opened 2024-02-29
ci: add dev task for testing build + cacert

mjnagel:
## Description

Adds a docker build task for basic local dev/CI validation.

Also adds a `cacert` task that runs the image and gets the cacert in base64 encoded format for use in the istio gateway.

## Related Issue

N/A

## Type of change

## Checklist before merging

### DIFF
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -10,6 +10,7 @@ variables:
 
 tasks:
   - name: build-and-publish
+    description: "Build and publish the multi-arch image"
     actions:
       - cmd: docker buildx build --push --platform linux/arm64/v8,linux/amd64 --tag ${IMAGE_NAME}:${VERSION} src
 
@@ -19,6 +20,7 @@ tasks:
       - cmd: docker build src -t uds-core-config:keycloak
 
   - name: dev-update-image
+    description: "Build the image and import locally into k3d"
     actions:
       - task: dev-build
       - cmd: |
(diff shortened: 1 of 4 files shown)


tasks.yaml
```diff
@@ -4,34 +4,53 @@ variables:
...
           PV=$(kubectl get pvc keycloak-themes -n keycloak -o jsonpath='{.spec.volumeName}')
           THEME_PATH=$(kubectl get pv $PV -o jsonpath="{.spec.hostPath.path}")
           docker cp src/theme k3d-uds-server-0:/$THEME_PATH
 
+  - name: cacert
+    description: "Get the CA cert value for the Istio Gateway"
+    actions:
+      # This is written to a file rather than printed because it is a massive value to copy out of the terminal
+      - cmd: |
+          docker run --rm --entrypoint sh ${IMAGE_NAME}:${VERSION} -c "cat /home/nonroot/authorized_certs.pem | base64 -w 0" > cacert.b64
```
jeff-mccoy, line 43: i think we could also do this in the script itself as a flag if that's less ugly? I dont have strong prefs either way.

rjferguson21 (approved):
LGTM - created https://github.com/defenseunicorns/uds-identity-config/issues/14 to track the need to document how this would be customized/updated.

tasks.yaml
mjnagel: Going to punt on that for now to get this across the line and cut a release of the image, but I think it would make things slightly nicer. Ideally no one besides us/developers cares about the command here, since they are just doing a `uds run` command.

PR #19, opened 2024-03-01
fix: account header username

UnicornChance:
## Description
Account Header was broken and now it isn't. 

Before: The welcome-page-script would attempt to get the users givenName and familyName, which it would do successfully, and then it uses this [reduce function](https://github.com/keycloak/keycloak/blob/release/23.0/themes/src/main/resources/theme/keycloak.v2/account/resources/welcome-page-scripts.js#L52) to combine those two names into a single name, with the caveat that it has to have a specific format specified in the messages_en.properties file. That value was not set correctly for the reduce function to use the parameters.

## Testing:
1. Setup uds-core with keycloak/authservice 
2. create admin user
3. go to `sso.uds.dev` and register a user
4. in admin console verify that users email address ( should now be able to get to the users account )
5. in the top right of the user account site, see the 'Full Name' beside the Sign Out button ( this should be the users full name )
6. with that local cluster still running, go to this repo and branch and run `uds run dev-theme` to copy in the new theme for keycloak
7. refreshing the users account should now show your users full name instead of the static text Full Name

...

## Type of change

## Checklist before merging


mjnagel (approved):
Took a while to get to this to test, LGTM!

PR #37, opened 2024-03-21
feat: inital smoke test

### LINKED ISSUE #27: Add basic smoke test for config changes
## New Feature
Need to add a/some basic smoke tests for making changes to the config so that it doesn't just break uds-core when it gets there.

A generic build of identity-config image and use that image to successfully deploy keycloak. 

Create task for building uds-core and then testing the new image with keycloak, update docs as well to represent that new task and how to use it.

### Additional Background Info
Further e2e/integration tests are not captured in this and will be implemented in uds-core repo.

UnicornChance:
## Description
Basic smoke test for uds-identity-config changes running against uds-core. This only tests that a change to the config will still deploy a healthy uds-core keycloak. More integration tests specific to registration and login and who knows what else are coming after this.

For the moment the git clone is a stopgap until we better understand what values need to be overridden when using the eventual k3d-core-slim image. Just want to get stuff in place as quickly as possible.

Add `--no-cache` flags to build process because of caching issues when building locally leading to false confidence in our processes. 

Tested locally and in CI ( personal repo). 
- `uds run uds-core-integration-test`

## Related Issue

Fixes #27 

## Type of change

## Checklist before merging


.github/workflows/test.yaml
```diff
@@ -27,3 +27,20 @@ jobs:
...
+    permissions:
+      pull-requests: read
+      contents: read
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Environment setup
+        uses: defenseunicorns/uds-common/.github/actions/setup@fc12e3a773580020a1d63e254525eab0f8b99fc8
```
rjferguson21, line 43: If you're going to use a SHA here definitely want to leave a comment for which version it came from. 

See: https://github.com/defenseunicorns/uds-package-gitlab/blob/7d31fb6b30c36d65f4c86b635372957e4f28ddc8/.github/workflows/test.yaml#L49

tasks.yaml
```diff
@@ -59,3 +59,11 @@ tasks:
           openssl req -new -key test.pem -out test.csr -config csr.conf -subj "/C=US/ST=Colorado/L=Colorado Springs/O=Defense Unicorns/CN=uds.dev"
           openssl x509 -req -in test.csr -CA ca.crt -CAkey ca.pem -CAcreateserial -out test.cer -days 365 -extensions v3_ext -extfile csr.conf
           zip src/authorized_certs ca.crt ca.pem test.cer test.csr test.pem
+
+  - name: uds-core-integration-test
+    description: "UDS Core integration test"
+    actions:
+      - cmd: uds deploy ghcr.io/defenseunicorns/packages/uds/bundles/k3d-core-istio-dev:0.16.1 --confirm
+      - task: dev-build
+      - cmd: git clone https://github.com/defenseunicorns/uds-core.git
```
rjferguson21, line 68: Probably want to pull a specific version of uds-core here; should be consistent with the base bundle version.

UnicornChance: @rjferguson21 add the version to be pulled as a renovate tracked variable. also was added in the cypress integration PR changes

rjferguson21:
Closing in favor of https://github.com/defenseunicorns/uds-identity-config/pull/42

PR #39, opened 2024-03-30
feat: zarf file

blancharda:
## Description

Occasionally, it may be useful to transport the identity-config image via zarf package. This PR Adds a simple `zarf.yaml` and corresponding UDS task for building it.

This will enable custom builds of this repository image to be more easily incorporated in UDS deployments, particularly in situations where the custom image may not be available at bundle create time.

## Type of change

### DIFF
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -2,7 +2,7 @@ variables:
   - name: VERSION
     description: "The image tag"
     # x-release-please-start-version
-    default: "0.3.7"
+    default: "0.3.6"
     # x-release-please-end
   - name: IMAGE_NAME
     description: "The repository + name for the published image"
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -13,16 +13,11 @@ tasks:
     description: "Build and publish the multi-arch image"
     actions:
       - cmd: docker buildx build --push --platform linux/arm64/v8,linux/amd64 --tag ${IMAGE_NAME}:${VERSION} src
-
-  - name: build-multi-arch
-    description: "Build the multi-arch image"
-    actions:
-      - cmd: docker buildx build --platform linux/arm64/v8,linux/amd64 --tag ${IMAGE_NAME}:${VERSION} src
   
   - name: build-zarf-pkg
     description: "Build the custom docker image and the zarf package for transporting it"
     actions:
-      - task: build-multi-arch
+      - task: docker build --tag ${IMAGE_NAME}:${VERSION} src
       - cmd: ./uds zarf package create . --set IDENTITY_CONFIG_IMG=${IMAGE_NAME}:${VERSION} --confirm
 
   - name: dev-build
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -17,7 +17,7 @@ tasks:
   - name: build-zarf-pkg
     description: "Build the custom docker image and the zarf package for transporting it"
     actions:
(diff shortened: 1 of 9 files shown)


zarf.yaml
```diff
@@ -0,0 +1,16 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/defenseunicorns/zarf/main/zarf.schema.json
```
Noxsios, line 1: should this use w/e version of zarf is `uds zarf version`?

blancharda: Solid maybe?
I definitely just copy/pasted this from somewhere... but version pinning seems like it might be useful.
Also feels like it would be easy to get out-of-date or cause some confusion based on whatever env specific version of UDS is being used 🤔 

Do we have a good renovate (or something) pattern for this anywhere?

blancharda: I also don't have a very good sense of how often the schema changes. If it's relatively infrequent maybe it's not worth worrying about since we would probably be encouraging staying up-to-date anyways..

tasks.yaml
```diff
@@ -14,6 +14,17 @@ tasks:
     actions:
       - cmd: docker buildx build --push --platform linux/arm64/v8,linux/amd64 --tag ${IMAGE_NAME}:${VERSION} src
 
+  - name: build-multi-arch
+    description: "Build the multi-arch image"
+    actions:
+      - cmd: docker buildx build --platform linux/arm64/v8,linux/amd64 --tag ${IMAGE_NAME}:${VERSION} src
```
blancharda, line 20: Does this realistically need to support multi-arch?

blancharda: Removed for now 🤷

zarf.yaml
mjnagel: Could switch this to the schema from uds-cli (`https://raw.githubusercontent.com/defenseunicorns/uds-cli/main/zarf.schema.json`) and pin to a specific uds tag + have it tracked by renovate if we want to be very careful here.

Separately - in uds-core we put schemas in our .vscode instead of adding at the top of the zarf file - https://github.com/defenseunicorns/uds-core/blob/main/.vscode/settings.json - maybe good to do that instead (and adopt the applicable renovate config to track that comment)?

mjnagel (commented):
No issues overall with this - seems useful for custom plugin builders. Could you document the new task in the readme + add a small note or two under docs/CUSTOMIZE.md to explain the zarf package piece?

zarf.yaml
blancharda: I think I added the renovate piece correctly, but could use a spot check before we resolve this one 👍

mjnagel (approved):

zarf.yaml
mjnagel: This should work - I think you technically only need [this single match](https://github.com/defenseunicorns/uds-core/blob/main/renovate.json#L69-L76) and might not even need the renovate comment piece...not sure it's being used in uds-core either.

blancharda: I ran it against a fork and it didn't seem to explode..
image

blancharda: Gotcha.
Good call.

The updates on the fork also didn't explode..
I'll push a simplified config to this PR.

mjnagel (approved):

mjnagel (approved):

PR #59, opened 2024-04-18
chore: update to integration test against main uds-core

rjferguson21:
## Description
Temporarily integration test against main for Keycloak 24 upgrade


UnicornChance (approved):

mjnagel (commented):
This will need to be 0.19.0 I believe - https://github.com/defenseunicorns/uds-identity-config/blob/main/bundles/uds-bundle.yaml#L25

mjnagel (approved):

UnicornChance (approved):

PR #65, opened 2024-04-23
fix: custom attribute plugin

UnicornChance:
## Description
By introducing an IDP, mattermost id was no longer being generated for those registration types. This happens because the Keycloak Authentication flow that is responsible for generating the mattermost id is tied to the registration form..which isnt accessed when using an IDP registration. 

By moving the mattermost id attribute generation out of the registration validation plugin and into it's an EventListener, we can toggle that on for all registration types. This also makes it simple for adding other attributes in the future ( although want to stay away from this as much as possible ).

There is also the alternative method of using a requiredAction, however consensus was to use the EventListener because of its extensibility. 
 
## Type of change

## Checklist before merging


UnicornChance:
Also was able to get an EventListenerProvider working. The following commit will be the inclusion of that piece for eyes.

Both options work and could be viable solutions. Comparing them:

---
EventListenerProviderFactory:

Pros:

- Flexibility: Allows you to listen to a wide range of events within Keycloak, not just user registration. You can react to events such as user login, logout, role assignment, etc.

- Fine-grained Control: You have precise control over the logic executed when the event occurs. This can be useful if you need to perform complex operations or interact with external systems during user registration.

Cons:

- Complexity: Implementing an EventListenerProviderFactory can be more complex than using a RequiredActionFactory, especially if you're not familiar with Keycloak's event system or the underlying technologies (like JBoss).
---
RequiredActionFactory:

Pros:

- Simplicity: Implementing a RequiredActionFactory is typically simpler and more straightforward than an EventListenerProviderFactory. It's specifically designed for extending user registration and authentication flows.

- Integration: Since it's part of the registration flow, using a RequiredActionFactory ensures that your attribute addition logic is seamlessly integrated into the registration process without requiring additional triggers or events.

- Scalability: This approach can be more scalable, especially if you're dealing with high volumes of user registrations. It's designed to work within the context of the authentication flow, which is already optimized for performance.

Cons:

- Limited Scope: The RequiredActionFactory is tied to user registration and may not be suitable if you need to perform actions in response to other events within Keycloak.
---
This specific implementation is fairly straight forward, so complexity isn't a bit issue at the moment but that could change in the future. Both of them are implementing code, so there is an ability to customize to our hearts content, however that comes with the caveat of maintenance.

jeff-mccoy (commented):
agreed this is a better overall solution and glad we chose this route over another requiredAction--very much the right call here. Question though--do we not have a unit test for this logic?

mjnagel (approved):

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/CustomEventListenerProvider.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public void onEvent(Event event) {
+
+        if (EventType.REGISTER.equals(event.getType())) {
+
+            RealmModel realm = this.model.getRealm(event.getRealmId());
+            UserModel newRegisteredUser = this.session.users().getUserById(realm, event.getUserId());
+
+            generateMattermostId(newRegisteredUser);
```
ericwyles, line 34: Question: Is there any need in here to check if the user already has a mattermostId before generating a new one? Since the id is date-based, if we were to ever regenerate it would change.

mjnagel: I think since we are only listening for the `REGISTER` event we should be safe? But it might not hurt to add a check regardless... curious on @UnicornChance's thoughts.

mjnagel (approved):

Issue #32, opened 2024-03-14
Set realm settings as environment variables


mjnagel:
### Is your feature request related to a problem? Please describe.

Currently if I want to change certain realm settings I would have to rebuild the realm or change it in the admin console at deploy time.

It would be ideal to have some realm items configurable at deploy time (currently `x509-cert-auth.ocsp-fail-open` feels like a good one to change).

### Describe the solution you'd like

Keycloak allows environment variables to be used in a realm import, see https://www.keycloak.org/server/importExport#_importing_a_realm_during_startup . This seems like an easy way to expose this as configurable (would need to be added to the keycloak helm chart to fully expose).

### Describe alternatives you've considered

End user can rebuild image with their custom realm.

### Additional context

We already use environment variables to handle the domain.

UnicornChance:
Potential value to look at is OTP.

PR #95, opened 2024-06-04
fix: expose environment variables

UnicornChance:
## Description
We've been asked by more than a couple people to expose some additional keycloak realm settings. This is that.

Added env vars for the following:
* Required Actions
  * Terms and Conditions
  * One Time Password ( OTP )
  * Email Verification
* Password Policy
* X509 OCSP Fail Open

In updating/testing the changes to the password policy, there was a couple changes necessary for the realms to handle the use of the `digits` in the `passwordPolicy`. While not strictly necessary a preemptive change to the cypress tests for user registration and login to update the passwords used was done as well.

#### There is an additional PR in the [uds-core repo](https://github.com/defenseunicorns/uds-core/pull/455) for setting the default uds-core variables for these changes.

The uds-core PR will need to be merged and uds-core released before CI jobs here will pass because of the dependency on the default env variables set in uds-core.

## Related Issue
Relates to #32 

## Type of change

## Checklist before merging

### DIFF
--- a/src/realm.json
+++ b/src/realm.json
@@ -438,7 +438,7 @@
     "requiredCredentials": [
         "password"
     ],
-    "passwordPolicy": "${REALM_PASSWORD_POLICY}:hashAlgorithm(pbkdf2-sha256) and forceExpiredPasswordChange(90) and specialChars(2) and passwordHistory(5) and length(12) and notUsername(undefined)",
+    "passwordPolicy": "${REALM_PASSWORD_POLICY:hashAlgorithm(pbkdf2-sha256) and forceExpiredPasswordChange(90) and specialChars(2) and passwordHistory(5) and length(12) and notUsername(undefined)}",
     "otpPolicyType": "totp",
     "otpPolicyAlgorithm": "HmacSHA1",
     "otpPolicyInitialCounter": 0,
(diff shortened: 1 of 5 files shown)


src/realm.json
```diff
@@ -437,7 +438,7 @@
     "requiredCredentials": [
         "password"
     ],
-    "passwordPolicy": "hashAlgorithm(pbkdf2-sha256) and forceExpiredPasswordChange(90) and specialChars(2) and passwordHistory(5) and length(12) and notUsername(undefined)",
+    "passwordPolicy": "${REALM_PASSWORD_POLICY}:hashAlgorithm(pbkdf2-sha256) and forceExpiredPasswordChange(90) and specialChars(2) and passwordHistory(5) and length(12) and notUsername(undefined)",
```
mjnagel, line 441: `}` in the wrong place?

rjferguson21 (approved):

rjferguson21 (approved):

src/realm.json
```diff
@@ -2799,7 +2814,7 @@
             "alias": "CONFIGURE_TOTP",
             "name": "Configure OTP",
             "providerId": "CONFIGURE_TOTP",
-            "enabled": true,
+            "enabled": "${REALM_OTP_ENABLED:false}",
```
mjnagel, line 2817: Would advocate for flip-flopping this back to `true`.

mjnagel (approved):

PR #98, opened 2024-06-06
fix: force sync to ensure copy occurs

rjferguson21:
## Description
Addresses an issue where existing files on disk are technically newer than files being copied in because of instances where initial installs have happened after the release of the `uds-identity-config` image.


mjnagel (approved):
LGTM - tested on an upgrade of core and seemed to behave as expected.

UnicornChance (approved):

PR #107, opened 2024-06-14
feat: basic group authz

rjferguson21:
## Description
* Adds authenticator that enforces client.attribute.groups membership
* Denies access if group does not exist
* Adds group authenticator to default realm configuration
* Adds group authenticator for IDP flow

This can be used with UDS Core by adding `attributes.groups` to  your client configuration in a Package:
```
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: httpbin-other
  namespace: authservice-test-app
spec:
  sso:
    - name: Demo SSO
      clientId: uds-core-httpbin
      attributes:
        groups: httpbin
      redirectUris:
        - "https://protected.uds.dev/login"
```        
Wanting to note I also had to add the Custom Authenticator to our default UDS Authentication flow as well as our configured Google SSO IDP as the post login flow:

## Type of change

## Checklist before merging

### DIFF
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
@@ -14,6 +14,8 @@ import org.keycloak.models.RealmModel;
 import org.keycloak.models.UserModel;
 import org.keycloak.sessions.AuthenticationSessionModel;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
  * Simple {@link Authenticator} that checks of a user is member of a given {@link GroupModel Group}.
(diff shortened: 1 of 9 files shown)


bburky:
Attributes can only be strings, right? An array of strings isn't possible?
It may make sense for someone to select multiple groups. Not sure what the best way to allow that would be.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -0,0 +1,163 @@
...
+        } else {
+            LOGGER.infof("%s No groups detected for %s client", logPrefix, client.getName());
+            success(context, user);
+        }
+    }
+
+    private Optional<GroupModel> getGroupByName(final String groupName, final RealmModel realm) {
+        return realm.getGroupsStream()
+            .filter(group -> group.getName().equals(groupName))
+            .findFirst();
```
bburky, line 74: How does this work with nested groups? Is this matching on the short group name or the full path?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -0,0 +1,163 @@
...
+     */
+    @Override
+    public void authenticate(final AuthenticationFlowContext context) {
+
+        RealmModel realm = context.getRealm();
+        UserModel user = context.getUser();
+        AuthenticationSessionModel authenticationSession = context.getAuthenticationSession();
+        ClientModel client = authenticationSession.getClient();
+        String clientId = client.getClientId();
+        String groupName = client.getAttribute("groups");
```
bburky, line 39: It make sense to namespace this or pick a slightly longer and more specific name? I wonder if `groups` may collide with a native keycloak attribute someday

bburky:
Can you check how the group sync code handles updating attributes via the `client-registrations` API? I assume if you change from `groups: foo` to `groups: bar` it will update the value. But if you delete `groups` entirely, will it remove the `groups` attribute in Keycloak? It's a PUT API that merges with the existing state I think, so it may not handle deletions well. I just want to verify.

rjferguson21:
Take away from our sync today:
* update the groups `client.attribute` field to be `uds.core.groups` (prefix to prevent collisions) instead of `groups`
* update the value of the groups attribute to be stored as JSON to avoid insanity with an alternative like a comma delimited string (where client names can contain commas)

In UDS Core
* Update the Package Custom Resource to accept an array, something like:
```
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: httpbin-other
  namespace: authservice-test-app
spec:
  sso:
    - name: Demo SSO
      clientId: uds-core-httpbin
      redirectUris:
        - "https://protected.uds.dev/login"
      enableAuthserviceSelector:
        app: httpbin
      groups:
        anyOf/allowed/required/requiredAny: #tbd key name
          - foo
          - bar
  ```

UnicornChance:
closing this PR in favor of #111 so that we can track P1 plugin lineage

PR #111, opened 2024-06-21
feat: group authorization plugin

UnicornChance:
## Description
* Adds authenticator that enforces client.attribute.groups membership
* Denies access if group does not exist
* Adds group authenticator to default realm configuration
* Adds group authenticator for IDP flow
* Unit Testing all the things

### Example Package definition to require group authentication for an application
```yaml
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: grafana
  namespace: {{ .Release.Namespace }}
spec:
  sso:
    - name: Grafana Dashboard
      clientId: uds-core-admin-grafana
      redirectUris:
        - "https://grafana.admin.{{ .Values.domain }}/login/generic_oauth"
      groups:
        anyOf:
          - /UDS Core/Admin
```

## Related Issue
This is related / has a sister PR in the [uds-core repo ](https://github.com/defenseunicorns/uds-core/pull/497)

## Type of change

## Checklist before merging

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,15 +31,12 @@ If the default realm, plugin, theme, truststore, or jars do not provide enough f
 
 This version upgrade brings in a new Authentication Flow for group authorization.
 
-If upgrading without a full redeploy of identity-config the following steps will be necessary to create and use group authorization:
+If upgrading without a full redeploy of keycloak the following steps will be necessary to create and use group authorization:
 1. In keycloak admin portal, in `UDS` realm, navigate to `Authentication` sidebar tab
 2. In `Authentication` tab add the `Authorization` flow to `UDS Authentication`, `UDS Registration`, `UDS Reset Credentials`
-   1. In each `Authentication` flow, `Add sub-flow`
-      1. `Name` -> `Authorization`
-      2. `Create`
-   2. Select the plus sign to `Add step` into the `Authorization` sub-flow
-      1. `Add` the `UDS Operator Group Authentication Validation`
-   * Make sure the `Authorization` flow is added at the base level and at the end of each `Authentication` flow
+   1. In each `Authentication` flow
+      1. `Add step` -> `UDS Operator Group Authentication Validation`
+      * Make sure that the step is at the base level and bottom of the Authentication flow
 3. Finally if using `SAML` IDP
    1. In the `Authentication` tab
       1. `Create Flow`
--- a/README.md
+++ b/README.md
@@ -23,3 +23,31 @@ This repo builds the UDS Identity (Keycloak) Config image used by UDS Identity. 
 ## Customizing UDS Identity Config
 
 If the default realm, plugin, theme, truststore, or jars do not provide enough functionality ( or provide too much functionality ), take a look at the [CUSTOMIZE.md](./docs/CUSTOMIZE.md) docs for making changes to the identity config.
+
+
+## Upgrading Identity Config
+
+### From v0.4.5 to v0.5.0
+
+This version upgrade brings in a new Authentication Flow for group authorization.
+
+If upgrading without a full redeploy of identity-config the following steps will be necessary to create and use group authorization:
+1. In keycloak admin portal, in `UDS` realm, navigate to `Authentication` sidebar tab
+2. In `Authentication` tab add the `Authorization` flow to `UDS Authentication`, `UDS Registration`, `UDS Reset Credentials`
+   1. In each `Authentication` flow, `Add sub-flow`
+      1. `Name` -> `Authorization`
+      2. `Create`
+   2. Select the plus sign to `Add step` into the `Authorization` sub-flow
+      1. `Add` the `UDS Operator Group Authentication Validation`
+   * Make sure the `Authorization` flow is added at the base level and at the end of each `Authentication` flow
+3. Finally if using `SAML` IDP
+   1. In the `Authentication` tab
+      1. `Create Flow`
+      2. `Name` -> `Authorization`
+      3. `Description` -> `UDS Operator Group Authentication Validation`
+      4. `Basic Flow`
+      5. `Create`
+      6. `Add execution`
+      7. `Add` the `UDS Operator Group Authentication Validation`
+   2. In the `Identity Providers` tab, select the `SAML` Provider
+      1. Add the `Authorization` flow to the `Post login flow` in the `Advanced settings` section
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
@@ -12,7 +12,6 @@ import org.keycloak.models.GroupModel;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.RealmModel;
 import org.keycloak.models.UserModel;
-import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
@@ -40,8 +39,8 @@ public class RequireGroupAuthenticator implements Authenticator {
             return;
         }
 
-        JsonNode groupsNode = parseGroupsAttribute(groupsAttribute);
-        if (groupsNode == null) {
+        Groups groups = parseGroupsAttribute(groupsAttribute);
+        if (groups == null) {
             LOGGER.warn("Failed to parse groups JSON");
             context.failure(AuthenticationFlowError.INVALID_CLIENT_SESSION);
             return;
@@ -51,8 +50,7 @@ public class RequireGroupAuthenticator implements Authenticator {
         boolean foundGroup = false;
         List<String> requiredGroups = new ArrayList<>();
 
-        for (JsonNode groupNode : groupsNode) {
-            String groupName = groupNode.asText();
+        for (String groupName : groups.anyOf) {
             requiredGroups.add(groupName);
             Optional<GroupModel> group = getGroupByPath(groupName, realm);
 
@@ -75,11 +73,11 @@ public class RequireGroupAuthenticator implements Authenticator {
         context.failure(AuthenticationFlowError.INVALID_CLIENT_SESSION);
     }
 
-    private JsonNode parseGroupsAttribute(String groupsAttribute) {
+    private Groups parseGroupsAttribute(String groupsAttribute) {
         ObjectMapper objectMapper = new ObjectMapper();
         try {
-            JsonNode groups = objectMapper.readValue(groupsAttribute, Groups.class).path("anyOf");
-            if(groups.size() == 0) {
+            Groups groups = objectMapper.readValue(groupsAttribute, Groups.class);
+            if(groups.anyOf.length == 0) {
                 LOGGER.errorf("Groups attribute does not contain a valid anyOf array");
                 return null;
             }
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
@@ -1,7 +1,6 @@
 package com.defenseunicorns.uds.keycloak.plugin.authentication;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.Arrays;
 import java.util.Optional;
 import org.jboss.logging.Logger;
 import org.keycloak.authentication.AuthenticationFlowContext;
@@ -48,10 +47,8 @@ public class RequireGroupAuthenticator implements Authenticator {
 
         RealmModel realm = context.getRealm();
         boolean foundGroup = false;
-        List<String> requiredGroups = new ArrayList<>();
 
         for (String groupName : groups.anyOf) {
-            requiredGroups.add(groupName);
             Optional<GroupModel> group = getGroupByPath(groupName, realm);
 
             if (group.isPresent()) {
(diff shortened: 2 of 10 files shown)


bburky:
`UDS Operator Group Authentication Validation config` is missing from the `UDS Registration` Keycloak authentication flow.
This allows bypassing group protection if the user does first-time keycloak user registration and directly logs into the app when finishing the registration flow.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        JsonNode groupsNode = parseGroupsAttribute(groupsAttribute);
+        if (groupsNode == null) {
+            logAndFail(context, "Failed to parse groups JSON", AuthenticationFlowError.INVALID_CLIENT_SESSION);
+            return;
+        }
+
+        handleGroupAuthentication(context, user, groupsNode);
+    }
```
bburky, line 47: I think I'd consider inlining `handleGroupAuthentication()` (it's only called on one place) and also inlining `logAndFail()`.

This makes it very clear that all code paths call `context.success()` or `context.failure()` before returning

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -0,0 +1,148 @@
...
+            LOGGER.errorf("Failed to parse groups JSON: %s", e.getMessage());
+            return null;
+        }
+    }
+
+    private void handleGroupAuthentication(AuthenticationFlowContext context, UserModel user, JsonNode groupsNode) {
+        if (!groupsNode.isArray() || groupsNode.size() == 0) {
+            logAndFail(context, "Groups attribute does not contain a valid anyOf array", AuthenticationFlowError.INVALID_CLIENT_SESSION);
+            return;
+        }
```
bburky, line 63: Move the `groupsNode.size() > 0` check inside `parseGroupsAttribute()`. (and you won't need to test `isArray()` if you parse into a class)

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @Override
+    public void action(final AuthenticationFlowContext authenticationFlowContext) {
+        // no implementation needed here
+    }
+
+    @Override
+    public boolean requiresUser() {
+        return false;
```
bburky, line 123:
```suggestion
        return true;
```

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -0,0 +1,148 @@
...
+
+    private void logAndFail(AuthenticationFlowContext context, String message, AuthenticationFlowError error) {
+        LOGGER.warn(message);
+        context.failure(error);
+    }
+
+    @Override
+    public void action(final AuthenticationFlowContext authenticationFlowContext) {
+        // no implementation needed here
+    }
```
bburky, line 119: Can we figure out when a `action()` on an `Authenticator` is called? What does "Called from a form action invocation."

If leaving this empty is correct, that's fine, but let's verify.

If we're sure that this should never be called, then maybe throw an `UnsupportedOperationException`.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
UnicornChance: moving the size check into `parseGroupsAttribute` does require passing in the context to that method for the context.failure. Something i was attempting to contain in the `authenticate` method. I'm indifferent if you feel strongly about that.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
UnicornChance: Based on ChatGPT, the action is called during the authentication flow after the user submits their credentials or interacts with the authentication form. It allows you to handle the user's input and decide the next step in the authentication process.

Essentially, it's a way to define a required action "like" actions that are specific to this authenticator. 

[Some example uses of the authenticator action method](https://github.com/thomasdarimont/keycloak-extension-playground/blob/2e16cb16b85d2ea61a1f6b095e3b01a13ec4af9d/dynamic-email-code-auth-extension/src/main/java/com/github/thomasdarimont/keycloak/auth/emailcode/EmailCodeAuthenticatorForm.java#L72)

So i am fairly confident that leaving this empty is the correct path forward.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        handleGroupAuthentication(context, user, groupsNode);
+    }
+
+    private JsonNode parseGroupsAttribute(String groupsAttribute) {
+        ObjectMapper objectMapper = new ObjectMapper();
+        try {
+            return objectMapper.readTree(groupsAttribute).path("anyOf");
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
UnicornChance: udpated

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -0,0 +1,148 @@
...
+                } else {
+                    foundGroup = true;
+                }
+            }
+        }
+
+        if (foundGroup) {
+            logAndFail(context, "User is not a member of the required group(s)", AuthenticationFlowError.INVALID_CLIENT_SESSION);
+        } else {
+            logAndFail(context, "Required group(s) do not exist in realm", AuthenticationFlowError.INVALID_CLIENT_SESSION);
```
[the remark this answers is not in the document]
UnicornChance: updated

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
UnicornChance: updated

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
UnicornChance: updated

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
bburky: There's already a log message and `return null` for a different error (invalid json). We can handle this similarly I think?

`return null` on error. check for null in `authenticate()` and do the `context.failure()` there

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
bburky: Hmmm ok. So may be like, if you had a "custom MFA" authenticator you could accept the OTP in a form dialog and this code might handle it? yeah, if that's accurate we don't need this

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
UnicornChance: updated

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -0,0 +1,144 @@
...
+                foundGroup = true;
+            }
+        }
+
+        if (foundGroup) {
+            LOGGER.warn("User is not a member of the required group(s)");
+            context.failure(AuthenticationFlowError.INVALID_CLIENT_SESSION);
+            return;
+        }
+        LOGGER.warnf("Required group(s) %s do not exist in realm", requiredGroups);
```
bburky, line 72: After switching to a `Groups` object, remove this ArrayList (and remove it's `import` too)
```suggestion

        for (String groupName : groups.anyOf) {
            requiredGroups.add(groupName);
            Optional<GroupModel> group = getGroupByPath(groupName, realm);

            if (group.isPresent()) {
                if (isMemberOfExactGroup(user, group.get())) {
                    LOGGER.infof("User %s is authorized for group %s", user.getUsername(), groupName);
                    context.success();
                    return;
                }
                foundGroup = true;
            }
        }

        if (foundGroup) {
            LOGGER.warn("User is not a member of the required group(s)");
            context.failure(AuthenticationFlowError.INVALID_CLIENT_SESSION);
            return;
        }
        LOGGER.warnf("Required group(s) %s do not exist in realm", Arrays.toString(groups.anyOf)));
```

UnicornChance: updated

bburky:
Group protection can also be bypassed using the "UDS Reset Credentials" authentication flow. I tested connecting a dev SMTP server (I used https://github.com/mailhog/MailHog) to Keycloak. And then attempted to access https://protected.uds.dev/ doing a password reset as part of the keycloak login flow.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -0,0 +1,148 @@
...
+
+    private String buildGroupPath(GroupModel group) {
+        StringBuilder path = new StringBuilder();
+        GroupModel currentGroup = group;
+        while (currentGroup != null) {
+            path.insert(0, "/" + currentGroup.getName());
+            currentGroup = currentGroup.getParent();
+        }
+        return path.toString();
+    }
```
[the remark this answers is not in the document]
UnicornChance: Unfortunately the `KC_SPI_GROUP_JPA_ESCAPE_SLASHES_IN_GROUP_PATH` is introduced in keycloak v25 which we've not upgraded to yet.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -0,0 +1,146 @@
...
+            .filter(group -> buildGroupPath(group).equals(groupPath))
+            .findFirst();
+    }
+
+    private String buildGroupPath(GroupModel group) {
+        StringBuilder path = new StringBuilder();
+        GroupModel currentGroup = group;
+        while (currentGroup != null) {
+            String groupName = currentGroup.getName();
+            if (groupName.contains("/")) {
```
rjferguson21, line 99:
```suggestion
            //  disallow slashes in group names to protect against Keycloak allowing unescaped slashes in group names until https://github.com/defenseunicorns/uds-identity-config/issues/118  
            if (groupName.contains("/")) {
```

UnicornChance: updated

src/realm.json
```diff
@@ -2250,13 +2251,41 @@
...
+                    "flowAlias": "Authorization",
+                    "userSetupAllowed": false,
+                    "autheticatorFlow": true
+                  }
+          
             ]
         },
+        {
+            "id": "7d53187d-13f4-4ec6-ade0-20b7d88d9aa4",
+            "alias": "Authorization",
```
mjnagel, line 2272: Not sure if it would belong in release here or in core (maybe both?) but would like to make sure we are tracking + documenting the manual upgrade addition of this flow to the realm (due to the current limitation around realm import and what pepr can manage).

mjnagel (approved):

README.md
```diff
@@ -23,3 +23,31 @@ This repo builds the UDS Identity (Keycloak) Config image used by UDS Identity.
...
 If the default realm, plugin, theme, truststore, or jars do not provide enough functionality ( or provide too much functionality ), take a look at the [CUSTOMIZE.md](./docs/CUSTOMIZE.md) docs for making changes to the identity config.
+
+
+## Upgrading Identity Config
+
+### From v0.4.5 to v0.5.0
+
+This version upgrade brings in a new Authentication Flow for group authorization.
+
+If upgrading without a full redeploy of identity-config the following steps will be necessary to create and use group authorization:
```
mjnagel, line 34:
```suggestion
If upgrading without a full redeploy of keycloak the following steps will be necessary to create and use group authorization:
```

nit, but I think technically redeploying keycloak/recreating the realm is the key here?

UnicornChance: updated

rjferguson21 (approved):

mjnagel (approved):

PR #112, opened 2024-06-21
fix: add group authz e2e tests

UnicornChance:
## Description
With the #111 Group Authz plugin created we now need some e2e tests. This PR creates two new e2e tests that show a user with the correct group to access the grafana dashboard and a user that does not have the correct group to not access the grafana dashboard.

## Related Issue
Relates to #111

## Type of change

## Checklist before merging


UnicornChance:
@mjnagel mentioned about updating grafana to be out of the box deployed with groups in uds-core since it already more or less requires group attribute within grafana. This would be reflected in the e2e tests being setup/configured here.

UnicornChance:
add some docs for how group auth works and something to the readme about uds docs site

tasks.yaml
```diff
@@ -102,16 +102,25 @@ tasks:
...
+      - cmd: |
+          sed -i '/^\s*sso:/,/^\s*-\s*name:/ { 
+            /groups:/d
+            /^\s*-\s*name:/a \ \ \ \ \ \ groups:\n\ \ \ \ \ \ \ \ anyOf:\n\ \ \ \ \ \ \ \ \ - \/UDS Core\/Admin
+          }' uds-core/src/grafana/chart/templates/uds-package.yaml
+
+      - cmd: |
+          cd uds-core
+          UDS_PKG=grafana uds run create-single-package
+          uds zarf package deploy build/zarf-package-uds-core-grafana-amd64.tar.zst --confirm --no-progress
```
mjnagel, line 123: Not to make us wait on another release, but we had briefly discussed maybe putting this in uds-core by default. Does that still seem like a possible route to clean this up (could be done after this PR)?

My only fear would be that someone could use overrides to change their grafana/neuvector group mappings and we would be restricting the client access at a different layer. We could make the group list templated with values as well, but they would still need to override two spots in that case.

src/test/cypress/support/commands.ts
```diff
@@ -71,12 +71,38 @@ Cypress.Commands.add("registrationPage", (formData: RegistrationFormData) => {
...
   // fill in user creds
   cy.get("#username").should("be.visible").type(username);
   cy.get("#password").should("be.visible").type(password);
 
   // click login button
   cy.get("#kc-login").should("be.visible").click();
 });
+
+/**
+ * Navigate to grafana ULR and verify redirected to sso.uds.dev
```
mjnagel, line 83:
```suggestion
 * Navigate to grafana URL and verify redirected to sso.uds.dev
```

docs/PLUGIN.md
```diff
@@ -6,7 +6,9 @@ weight: 2
...
 
-One nuanced auth flow is the creation of a Mattermost ID attribute for users. [CustomEventListener](https://github.com/defenseunicorns/uds-identity-config/blob/main/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/CustomEventListenerProvider.java) is responsible for generating the unique ID.
+One of the plugin capabilities is [group authentication](https://github.com/defenseunicorns/uds-identity-config/blob/main/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java). This capability allows for defining a keycloak group membership that is required to access an application. There is [an e2e test](https://github.com/defenseunicorns/uds-identity-config/blob/main/src/test/cypress/e2e/group-authz.cy.ts) that demonstrates grafana being deployed and requiring that users accessing its dashboard be members of the `/UDS Core/Admin` group. There is [additional documentation](https://github.com/defenseunicorns/uds-core/blob/main/docs/configuration/uds-operator.md?plain=1#L23-L26) for configuring an application for group authentication in uds-core.
```
mjnagel, line 9: nit: may want to pin the uds-core link to a sha or tag rather than main in case the docs line numbers change.

mjnagel (commented):
Small comments, overall LGTM.

src/test/cypress/support/commands.ts
UnicornChance: updated

docs/PLUGIN.md
UnicornChance: good call, updated

tasks.yaml
UnicornChance: yeah we've talked about this a few times and i don't think we've actually attempted to try it. I think it would be a good idea to look into this.

mjnagel (approved):

tasks.yaml
mjnagel: I'll leave that for a follow-on for sure, but worth exploring. That override scenario is the only thing that gives me pause, but we might be able to handle nicely or just enforce "you must use our groups".

UnicornChance: https://github.com/defenseunicorns/uds-identity-config/issues/137

PR #136, opened 2024-07-09
fix: x509 bug fixes

### LINKED ISSUE #125: x509 issues
### First Issue:
Occasionally when using a CAC, the pop up that displays user info from the CAC is empty string.

### Steps to reproduce
1. new registration with saml and new pki
2. login with google
3. next page should be a DoD PKI Detected and display "No Certificate"

### Expected result
Should always show the users info instead of "No Certificate"

### Additional Context
After initial login the attribute hasn't been added to user yet and so it shows "No Certificate", however if you log out and back in it shows correctly.

image

### Second issue:
Another issue is that the registration CaC flow shows an email that doesnt mean anything to humans. Should look into extracting different user data from cac in the them, registry.ftl cacIdentity field.

image

### Third issue:
with cac present and forgot password the email is autofilled to be something from the cac that isn't relevant and should be users email if autofiled, although we probably dont have that so nothing should be autofilled. 

image

### Fourth issue:
password reset page shows help@dsop.mil
Screenshot from 2024-07-08 11-43-27

## Completion
Have @bburky review code changes.

UnicornChance:
## Description
Lots of fixes for findings from @bburky review of our authentication flows. Most of this is theme changes and what is displayed to the end user. There is some added logic handling in the plugin to extract a users SubjectDN for display back to the user instead of displaying the CN ( just better information ). 

Directly fixed things in this PR:

1. The registration CaC flow shows an email that doesnt mean anything to humans, swapped this to using the x509 SubjectDN instead of the CN.
2. with cac present and forgot password the email is autofilled to be something from the cac that isn't relevant. Brought in the default keycloak theme file so that we can control this autofill feature ( ie turning it off ).
3. password reset page uses dsop@mil email address which is controlled from the messages_en.properties file, updated that to not reference an email. If an email is required here, there would need to be a custom theme implemented for supplying that message with the correct email.
4. Using X509 and SAML together resulted in the x509-login-info page showing a `[No Certificate]` despite a certificate being present. This was caused by keycloak default behavior not properly setting the subjectDN, added some logic to parse the subjectDN in our Update X509 plugin and add that to the form when keycloak default behavior doesn't work properly.

## Related Issue

Fixes #125 

## Type of change

## Checklist before merging


bburky (changes requested):
Tested locally, looks good overall.

One minor html escaping issue in comments.

Also, do we need the old `cacIdentity`? Is it unused now?

src/theme/login/login-reset-password.ftl
```diff
@@ -0,0 +1,39 @@
...
+        <form id="kc-reset-password-form" class="${properties.kcFormClass!}" action="${url.loginAction}" method="post">
+            <div class="${properties.kcFormGroupClass!}">
+                <div class="${properties.kcLabelWrapperClass!}">
+                    <label for="username" class="${properties.kcLabelClass!}"><#if !realm.loginWithEmailAllowed>${msg("username")}<#elseif !realm.registrationEmailAsUsername>${msg("usernameOrEmail")}<#else>${msg("email")}</#if></label>
+                </div>
+                <div class="${properties.kcInputWrapperClass!}">
+                    <input type="text" id="username" name="username" class="${properties.kcInputClass!}" autofocus aria-invalid="<#if messagesPerField.existsError('username')>true</#if>" dir="ltr"/>
+                    <#if messagesPerField.existsError('username')>
+                        <span id="input-error-username" class="${properties.kcInputErrorMessageClass!}" aria-live="polite">
+                                    ${kcSanitize(messagesPerField.get('username'))?no_esc}
```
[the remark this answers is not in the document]
UnicornChance: updated

UnicornChance:
the `cacIdentity` form attribute is there only for conditional rendering of x509 related messages to the user. It doesn't have a functional use outside of that. I could see use renaming that to be more applicable but removing it would remove necessary x509 information.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/X509Tools.java
```diff
@@ -213,4 +213,54 @@ private static Object getX509Identity(final KeycloakSession session, final HttpR
...
+     * @param Keycloak form context http request
+     * @param Keycloak form keycloak session provider certificate lookup
+     * @return String
+     */
+    public static String getX509SubjectDN(KeycloakSession session, HttpRequest httpRequest,
+            X509ClientCertificateLookup provider) {
+
+                try {
+                    X509Certificate[] certs = provider.getCertificateChain(httpRequest);
+                    return certs[0].getSubjectX500Principal().getName();
```
bburky, line 258: Wrap this with `if (certs.length >= 0)` (and let the final `return null` run if empty)

UnicornChance: updated

bburky (approved):

rjferguson21:
Tested with the merged user profile stuff and everything looks good.  :+1:

rjferguson21 (approved):

PR #138, opened 2024-07-10
fix: escape client name

### LINKED ISSUE #128: XSS via client name in login page
`client.name` is rendered into the webpage without escaping:
https://github.com/defenseunicorns/uds-identity-config/blob/v0.5.0/src/theme/login/template.ftl#L57
XSS is possible through the display name of a Keycloak client.

Even with client self-registration via Package CR, I think we still consider creating clients to be an admin action, so this has limited security impact (an admin can XSS users, but non-admins shouldn't be able exploit this)

UnicornChance:
## Description
Currently a client name is not escaped allowing for a potential XSS. Wrap client name in Keycloak's `kcSanitize` function to make sure that value is escaped properly.

## Related Issue

Fixes #128 

## Type of change

## Checklist before merging

### DIFF
--- a/src/theme/login/template.ftl
+++ b/src/theme/login/template.ftl
@@ -50,7 +50,7 @@ ${msg("loginTitle",(realm.displayName!''))}
                                     </h2>
                                 <#else>
                                     <h2 class="client-unique-name">
-                                        ${client.name?no_esc}
+                                        ${kcSanitize(client.name)?no_esc}
                                     </h2>
                                 </#if>
                             <#else>


bburky:
Apparently `kcSanitize` is `KeycloakSanitizerMethod`:
https://www.keycloak.org/docs-api/25.0.1/javadocs/org/keycloak/theme/KeycloakSanitizerMethod.html

This actually does still allow html, but enforces an allowlist based policy:
https://github.com/keycloak/keycloak/blob/25.0.1/services/src/main/java/org/keycloak/theme/KeycloakSanitizerPolicy.java

I think this is probably fine. I verified it blocks script tags with manual testing, and the HTML allowlist seems relatively narrow and shouldn't allow scripts from other elements or attributes. This should block XSS.

I don't think we actually need to allow HTML at all, but it seems fine I guess. Could still be slightly surprising behavior though if someone tries to put `<` or `>` characters into the name. But that seems fairly unlikely.

rjferguson21 (approved):

PR #144, opened 2024-07-15
fix: registration validation

### LINKED ISSUE #126: Registration Validation
### Description
We need to do the custom RegistrationValidation registration-validation-action  for SAML registration flows too.

In  RegistrationValidation registration-validation-action we at a miniumum should validate all user.attributes.* in the form data (including unexpected attributes), but could validate all formdata fields

### Completion
**When completed have @bburky review changes**

### Severity/Priority
High Priority

UnicornChance:
## Description
Revert previous merged PR and implement keycloak builtin functionality for managing user attributes.

[Documentation for this builtin functionality](https://www.keycloak.org/docs/25.0.1/server_admin/#user-profile)

## Related Issue

Fixes #126 

related previous PR #141 

## Type of change

## Checklist before merging


bburky (changes requested):
Cool, this new feature seems to work. I tested that users cannot set `mattermostid` or `user.attributes.mattermostid` during account registration.

However, this is currently hiding users' bound CAC id attribute `usercertificate` in the admin UI. (surprisingly this unmanaged attribute can be set anyway via Java code, but is just invisible in the admin UI.) Please add `usercertificate` ad an additional admin-editable-only attribute.

Also, please set "Realm settings" → "General" → "Unmanaged Attributes" to "Only administrators can write". In case there are any more attributes we've missed, this lets an admin still access them in the UI.

I'm going to do a little bit more testing, but this approach seems sane and actually has a better admin UI.

bburky:
Neat, the "required" setting on managed attributes triggers a registration dialog (oddly _different_ than our normal registration dialog, I wonder if we could move away from some of our custom theme forms in the future). Ideally this new dialog would match our normal registration form a bit better ("notes" needs an explanation for example, or just not be shown at all).

My only real issue with this form is that SAML users can edit their name and email on it. If they pick a different email their email becomes unverified, which is good. But ideally they can't edit their username and email at all. If we set `username` and `email` to "Who can edit?" and "Admin" only (disable "user") it does disappear from the form (it becomes uneditable). However... email probably _should_ be editable for non-SAML Keycloak installations.

Anyways, you're not currently setting any fields to required so none of that applies at the moment. But we probably should make an issue for "admin can pick between user/pass, SAML, and x509 login" (disabling the ones we don't want) with a related configuration of "require email verification" and better handling of the `email` JWT claim.

For now, please set `username` to "who can edit?" and disable "user" (allow "admin" edit only). Skip email I think.

bburky:
Is there any chance we can make all of our `RegistrationValidation` checks into (custom?) validations of the managed attributes? Things like `mattermostUsernameValidation()` are still missing in the SAML registration flow.

bburky:
For the `notes` attribute, please set the "Display name" to "Access Request Notes", and add the annotation `inputType` set to `textarea`.

For the `organization` attribute, please set the "Display name" to "Unit, Organization or Company Name".

Is there any reason to pick an "attribute group"? It actually shows in the UI as "User Attributes" which can sometimes be shown to users. Picking "None" puts them in no group and the "User Attributes" heading never shows in the UI. It looks like you created this group? is it needed?

(these don't matter too much yet, but if we set anything to required they customize the SAML review profile form)

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/RegistrationValidation.java
```diff
@@ -142,20 +119,6 @@ public void validate(final ValidationContext context) {
...
-
-        // Require a rank
-        if (Validation.isBlank(formData.getFirst("user.attributes.rank"))) {
-            errors.add(new FormMessage("user.attributes.rank", "Please specify your rank or choose n/a."));
-        }
-
-        // Require an organization
-        if (Validation.isBlank(formData.getFirst("user.attributes.organization"))) {
-            errors.add(new FormMessage("user.attributes.organization", "Please specify your organization."));
-        }
```
bburky, line 158: These were previously required, so this is change for these to be optional. Maybe this is okay, but I was hoping we could have these configurable (I guess making them configurable can be later work). Can we check with current users (Navy I guess?) if it's ok to make these optional

bburky:
`mattermostUsernameValidation()` can be replaced with a "pattern" validation of `[A-Za-z][A-Za-z0-9-_.]{0,31}`. You'll need to add a new error message string to the `messages_en.properties` file and provide the key a "Error message key".

This does introduce a problem though: the current SAML username is the email address, which actually violates this policy. My username on Mattermost _does_ seem to be set to `blake-defenseunicorns.com` and is unconfigurable. This validation probably is the correct thing to do with the current mattermost config. Unsure what the right fix is, allow email? Change the user mapping for mattermost? leave it?

Set first name and last name to required, that covers their validation.

Duplicate email checking and a email not blank seem to be built in keycloak checks and I think can be removed.

I don't see how to do the "Sorry, this CAC seems to already be registered." check without custom code, but as far as I can tell Keycloak is actually enforcing this silently anyway. (it silently doesn't set `usercertificate` if there's a duplicate). Probably fine to have this as a Java validation on the manual user registration form only and have it silently dropped on the SAML flow I guess?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/RegistrationValidation.java
UnicornChance: By moving to the user managed attributes we can configure these fields to be required or not via the realm.json. Which is actually applicable to other fields in this RegistrationValidation.java class, the checks here are mostly redundant if only checking that the field is blank.

bburky (approved):

rjferguson21 (approved):

PR #165, opened 2024-07-25
fix: add new client scope

UnicornChance:
## Description
Create an optional Client Scope for including only the child group name and excluding all hierarchical group paths. 

The new client scope is called `bare-groups` and it creates a token claim called `bare_groups`. This scope is optional, meaning it will not come default with clients, we still want to encourage the use of the full.path group called `groups` that is in the `profile` client scope. 

There is a potential risk if using only the new `bare_groups` for user authorization because there can technically be two different child groups called `ADMIN` and by removing the parent group structure those two child groups would be identical. An easy solution for this is to implement the use of [group authorization](https://uds.defenseunicorns.com/core/configuration/uds-operator/#package) that uses the `group` full.path to verify a user has access to the given application.

## Related Issue

Fixes [#603](https://github.com/defenseunicorns/uds-core/issues/603)

## Type of change

## Checklist before merging

### DIFF
--- a/src/realm.json
+++ b/src/realm.json
@@ -1252,6 +1252,35 @@
             ]
         },
         {
+            "id": "1962f32a-2c09-4ce0-8677-6f6776768d24",
+            "name": "bare-groups",
+            "description": "A client scope providing only the simplified, non-hierarchical group names.",
+            "protocol": "openid-connect",
+            "attributes": {
+                "include.in.token.scope": "true",
+                "display.on.consent.screen": "true",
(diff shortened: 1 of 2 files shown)


rjferguson21 (approved):
Looks good.

rjferguson21:
In flight discussion happening in https://github.com/defenseunicorns/uds-core/issues/603

UnicornChance:
Closing this PR in favor of #175

PR #171, opened 2024-07-30
feat: creation of event listener

UnicornChance:
## Description
New Event Listener for keycloak events to log in json format for easier consumption by grafana or other observability applications.

Moved existing event listener into a director for event listeners. Created new json log event listener.

Json-log-event-listener-grafana-example

Also removed the jboss-logging event listener from realms.

## Related Issue

Fixes [#522](https://github.com/defenseunicorns/uds-core/issues/522)

## Type of change

## Checklist before merging


src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/eventListeners/JSONLogEventListenerProvider.java
```diff
@@ -0,0 +1,71 @@
...
+                .withZone(ZoneOffset.UTC)
+                .format(instant);
+
+        // Create a new ObjectNode for ordered fields
+        ObjectNode jsonMessage = objectMapper.createObjectNode();
+
+        // Add fields in the desired order
+        jsonMessage.put("timestamp", formattedTime);
+        jsonMessage.put("loggerName", "uds.keycloak.plugin.eventListeners.JSONLogEventListenerProvider");
+        jsonMessage.put("level", "DEBUG");
```
rjferguson21, line 56: I'm open to other opinions here but IMO we should just drop the log level.  I think it would only be confusing for folks since we're not actually using the logger.

UnicornChance: yeah i think that's fair.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/eventListeners/JSONLogEventListenerProvider.java
```diff
@@ -0,0 +1,70 @@
...
+        while (fields.hasNext()) {
+            Entry<String, JsonNode> field = fields.next();
+            jsonMessage.set(field.getKey(), field.getValue());
+        }
+
+        return jsonMessage;
+    }
+
+    private JsonNode convertAdminEventToJson(AdminEvent adminEvent) {
+        return objectMapper.valueToTree(adminEvent);
```
rjferguson21, line 67: Are we not logging the same metadata for Admin events?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/eventListeners/JSONLogEventListenerProvider.java
```diff
@@ -0,0 +1,70 @@
...
+        String formattedTime = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSSSSSSSX")
+                .withZone(ZoneOffset.UTC)
+                .format(instant);
+
+        // Create a new ObjectNode for ordered fields
+        ObjectNode jsonMessage = objectMapper.createObjectNode();
+
+        // Add fields in the desired order
+        jsonMessage.put("timestamp", formattedTime);
+        jsonMessage.put("loggerName", "uds.keycloak.plugin.eventListeners.JSONLogEventListenerProvider");
```
rjferguson21, line 55: Can we add another property to both types of events,  but would distinguish which type they where? 

So maybe eventType: "USER" / "ADMIN"?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/eventListeners/JSONLogEventListenerProvider.java
UnicornChance: similar to your other comment was leaving the admin events untouched as a differentiator but definitely see the use case for them being formatted similarly. Updated.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/eventListeners/JSONLogEventListenerProvider.java
UnicornChance: similar to your other comment was leaving the admin events untouched as a differentiator but definitely see the use case for them being formatted similarly. Updated.

PR #184, opened 2024-08-05
chore: update docs

UnicornChance:
## Description
Need to update docs for all the new plugins that have been created over the past weeks. 

Critically, i'm not sure if we should document all of them the way i have it now. For example the `CustomGroupPathMapper` maybe should be undocumented to remove confusion and limit its use. 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Checklist before merging

- [x] Test, docs, adr added or updated as needed
- [x] [Contributor Guide Steps](https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md)(https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md#submitting-a-pull-request) followed

PR #185, opened 2024-08-05
feat: new aws mapper

### LINKED ISSUE #181: feat: add Custom Group Mapper
### Is your feature request related to a problem? 
When leveraging Keycloak as a SAML IdP for some AWS services, there isn't a great way to do fine grained permissions due to how AWS requires groups to be passed in and what our IdP is able to do.

- **Given** a Keycloak SAML client is created for an AWS managed service
- **When** a user authenticates leveraging that client 
- **Then** I can pass in a user's group attributes in a way that AWS can leverage them

### Describe alternatives you've considered
- Open a support case with AWS requesting additional SAML capabilities on the SP side
- Create multiple clients per granular capability / AWS accounts and hard code a group attribute to each

### Additional context
[Link to required group format](https://docs.aws.amazon.com/appstream2/latest/developerguide/application-entitlements-saml.html#:~:text=where%20%7BTagKey%7D%20is%20one%20of%20the%20following%20attributes%3A) "group1:group2:group3"

This will reduce complexity/ manual configuration and improve user experience. At our current small scale, we need six clients and urls for users to interact with. With this capability enabled we can reduce that by 60% (one per account/managed service).

For future scale, this exponentially provides a much better user experience. I am unsure if this would be useful in other contexts.

UnicornChance:
## Description
New Client Scope mapper for formatting users groups to be usable for AWS AppStream. 

New Optional Client Scope that creates the custom groups claim called `aws-groups` in the format requested. Unit tests covering large amounts of new code. Image below demonstrating the new claim.

aws-group-mapper

Moved the groups claim variable to common to be used via both group mappers we are now maintaining.

[follow on PR](https://github.com/defenseunicorns/uds-identity-config/pull/184) for more documentation, creating separately so this PR doesn't get caught up in details around what that looks like. 

## Related Issue

Fixes #181 
Relates to #184 ( docs update )

## Type of change

## Checklist before merging


ntwkninja (approved):
lgtm

rjferguson21 (approved):

src/realm.json
```diff
@@ -1633,6 +1633,25 @@
...
+            },
+            "protocolMappers": [
+                {
+                    "id": "a87cb1d0-214b-4e92-8b93-562efb69f36d",
+                    "name": "aws-groups",
+                    "protocol": "saml",
+                    "protocolMapper": "aws-saml-group-mapper",
+                    "consentRequired": false
+                }
+            ]
```
mjnagel, line 1654: If we do want to pull this into the baseline (open question I think) I would lean towards not adding this into the realm given that we don't expect this default one to be used by known asks for it.

mjnagel (approved):
LGTM, noting the follow-on PR to document this and the other features to ensure they are more discoverable.

mjnagel (commented):
Tiny comment on the PR title/merge commit - this might be a feat or chore at least? Doesn't feel like a fix.

PR #271, opened 2024-12-02
feat: expose timeouts

UnicornChance:
## Description

A request was made so that we can expose the keycloak timeouts to be overridden for demo purposes. Update exposes access token lifespan and sso session idle timeout. To be clear, these values still maintain the same defaults and uds-core will not be updating the values we use out of the box in our bundles. 

Add the exposed environment variables in the example customization doc as well. 

## Type of change

## Checklist before merging


noahpb (dismissed):

src/realm.json
```diff
@@ -5,8 +5,8 @@
     "displayNameHtml": "Identity Service",
     "notBefore": 0,
     "defaultSignatureAlgorithm": "RS256",
-    "revokeRefreshToken": false,
-    "refreshTokenMaxReuse": 0,
+    "revokeRefreshToken": true,
+    "refreshTokenMaxReuse": 1,
```
mjnagel, line 9: Should we document this in our realm/identity config upgrade doc (https://github.com/defenseunicorns/uds-identity-config/blob/main/docs/reference/UDS%20Core/IdAM/upgrading-versions.md)?

Sidenote I also realized our docs links on the readme are currently broken ([here](https://github.com/defenseunicorns/uds-identity-config/blob/main/README.md#customizing-uds-identity-config) and [here](https://github.com/defenseunicorns/uds-identity-config/blob/main/README.md#upgrading-identity-config)).

Issue #280, opened 2024-12-04
Allow disabling username/password authentication for master realm


mjnagel:
Identify a path forward for completely disabling user/pass based auth on the master realm (i.e. if I want to enforce CAC only authentication).

The outcome of this should be one of the following:
- toggles + documentation for disabling user/pass
- guide on how to disable user/pass in a custom "fork" of identity config (and POC of this)

UnicornChance:
created documentation for this. We discussed actually addressing this a configuration and while its viable with master realm imports or terraform provider. Something we should look at doing in the new year though so that we can lock down the admin portal out of the box for uds core deployments.

UnicornChance:
Closing this issue in favor of #298

Issue #152, opened 2024-07-17
deploy time configuration for identity type


UnicornChance:
### Describe the solution you'd like
At deploy time a deployment could be configured to pick between user/pass, SAML, and/or x509 Login. Most likely this would be an opt out implementation, and the resulting configuration would alter the way validations are done. This would also allow for better handling of the email JWT claim.

UnicornChance:
OBE from work in #282

Issue #148, opened 2024-07-16
Required MFA has some edge cases


bburky:
## Steps to reproduce
1. Create a Keycloak account via SAML (and probably via x509 too)
2. Skip adding a password
3. Go to the user profile and set a password https://sso.uds.dev/realms/uds/account/#/security/signingin
4. Observe that the user was created in keycloak and the user can log in with user and password only

Or

1. Create a new user with a password
2. Close the browser window when it prompts you to set MFA
3. Observe that the user was created in keycloak and the user can log in with user and password only

In both cases MFA will be enforced on next login with password, but that still is technically a single factor login. The user may already have permissions granted by an admin (added to Keycloak groups) while it is in this single factor state. 

### Expected result

whenever a password is set, MFA should also be required. If MFA is somehow skipped, the user account shouldn't be created at all during new user registration. When setting a first password on an existing user, require MFA before saving the password to the existing account.

### Actual Result

User accounts can be created with a password only. MFA is enforced on next login, but this first password login is actually a single factor login.

UnicornChance:
Configuring keycloak to not create a user unless successfully completing the entire registration form and MFA setup seems like it would be some kind of custom plugin? Maybe you have some more insight into how this would work.

I think the process of first login expecting an MFA to be configured if somehow skipped during registration is the standard industry approach and would lead to a single login where a SFA was necessary. 

This is also somewhat similar but also kinda conflicting with [this other MFA issue](https://github.com/defenseunicorns/uds-identity-config/issues/117).

Just looking for a good path forward.

PR #282, opened 2024-12-09
feat: configurable authentication flows

### LINKED ISSUE #279: Allow disabling username/password authentication for UDS realm
Identify a path forward for completely disabling user/pass based auth on the uds realm (i.e. if I want to enforce CAC only authentication).

The outcome of this should be one of the following:
- toggles + documentation for disabling user/pass
- guide on how to disable user/pass in a custom "fork" of identity config (and POC of this)

UnicornChance:
## Description

**There are no breaking changes in this PR.**

Everything that is now configurable has a default value that is set to the original value. In the future we may add breaking changes to focus on a specific flow, but at this time we will maintain all three auth flows as default.

### Back Story
At this time uds-core + identity-config allows for three different authentication flows: X509, Username Password, Social ( Google SSO ). Having three different authentication flows is often not desired for environments, typically only one or two of these flows is desired, i.e. x509 only, or social and x509. 

Keycloak, by default, does not have simple configurations for turning off these different flows individually. Specifically Username Password can be a major pain to turn off in an environment, taking several different settings to entirely disable it. 

These changes allow for the configuration of realm and theme to control auth flows from bundle environment variables, [a related PR is up in uds-core](https://github.com/defenseunicorns/uds-core/pull/1102) for these changes to work. The bundle variables are put into a secret, which is then loaded into the statefulset for the identity config init container, which runs the `sync.sh` and adds those environment variables to the theme.properties file, the realm.json ( which configures the realm settings ) grabs the environment variables by itself.

## Testing
Configuring the different auth flows as described in the new doc `authentication-flows.md` and testing that a user can only register / authenticate via that method. There is a few different ways to do this, configure via bundle and redeploy, update bundle overrides and cycle keycloak things.

**Make sure you pull in the corresponding [uds-core changes / branch](https://github.com/defenseunicorns/uds-core/pull/1102) so that the keycloak secret values can be mapped into the theme.properties**

## Explanation of changes
In the theme, I needed to control how the theme would present different pieces, x509 block, username password, social, register button etc. While conditionally rendering the theme does not provide real security it does improve user experience and attempts to guide the user to the desired outcome. Pairing these changes with real security controls was mostly necessary for UX purposes.

Plugin logic, when registering with a x509, the user still needs to go through the register form. Previously the password and password confirm fields were still present on that form, although they weren't required. The x509 plugin logic required that the password and password confirm fields were returned from the form even if they were empty. By conditionally removing those fields it would break the plugin because the password/confirm fields were null rather than empty.

Realm changes, as mentioned previously the theme changes are only a band-aid fix and a motivated person could still potentially set a password in a non username password environment or reset their password or something to that effect. These configurations are meant to be paired with the theme configurations to actually apply the security measures.

## Related Issue

Fixes #279 #152 

Relates to #280 #148 

## Type of change

## Checklist before merging


mjnagel (commented):
Haven't fully deployed yet, but had a few small comments.

src/test/cypress/realm.json
```diff
@@ -2683,7 +2691,7 @@
             "alias": "CONFIGURE_TOTP",
             "name": "Configure OTP",
             "providerId": "CONFIGURE_TOTP",
-            "enabled": "${REALM_OTP_ENABLED:true}",
+            "enabled": "${REALM_OTP_ENABLED:false}",
```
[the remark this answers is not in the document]
UnicornChance: Yeah, this was intentional. This is the testing realm.json, I made a few changes to the realm to make it as similar to the main realm.json as possibly however I've never setup OTP testing so disabling that by default was required for successful cypress testing.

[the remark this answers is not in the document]
mjnagel: Ah duh, I didn't look at the name carefully.

UnicornChance:
introduced a breaking change with the renaming of `OTP_ENABLED`, will fix and align other envs with that syntax.

src/realm.json
```diff
@@ -2910,7 +2918,7 @@
             "alias": "CONFIGURE_TOTP",
             "name": "Configure OTP",
             "providerId": "CONFIGURE_TOTP",
-            "enabled": "${REALM_OTP_ENABLED:true}",
+            "enabled": "${OTP_ENABLED:true}",
```
mjnagel, line 2921: Did this still end up being a breaking change due to the removal of the `REALM_` prefix? I think that's okay since this only gets read in/evaluated at install time, just wanted to call it out and ask.

mjnagel: Actually in thinking this through the change might be a bit more problematic, I dropped a comment on the core side - https://github.com/defenseunicorns/uds-core/pull/1102/files#r1884100380

src/sync.sh
```diff
@@ -6,6 +6,9 @@ set -e
 
 echo "Syncing customizations to Keycloak"
 
+# Dump environment variables for debugging
+# printenv > /opt/keycloak/data/env_dump.txt
+
```
mjnagel, line 11: nit: clean up the commented out code

docs/reference/UDS Core/IdAM/authentication-flows.md
```diff
@@ -0,0 +1,70 @@
...
+
+| Setting | Description | Options |
+| - | - | - |
+| [USERNAME_PASSWORD_AUTH_ENABLED](https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/chart/values.yaml#L30) | Toggle on/off the Username and Password Authentication flow. When disabled there will be no username password login, password / password confirm registration fields, no credential reset, and no update password options available. | `true`(default), `false` |
+| [X509_AUTH_ENABLED](https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/chart/values.yaml#L31) | Toggle on/off X509 (CAC) Authentication flow. | `true`(default), `false` |
+| [SOCIAL_AUTH_ENABLED](https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/chart/values.yaml#L32) | Toggle on/off Social (Google SSO, Azure AD, etc. ) Authentication flows.| `true`(default), `false` |
+
+These three variables handle the complexities of configuring the following environment variables, which are responsible for both visual (theme) and security (realm). The following variables are not exposed for overriding.
```
mjnagel, line 24: I'm wondering if we should even have most of the doc after this (with the exception of the manual config)? If it's not exposed them I'm unsure if it's really providing value for the end user - but maybe it's more of a dev doc for our reference of all the toggles?

mjnagel (approved):
LGTM overall - few small comments. I validated the flow toggles with core and didn't encounter any issues.

docs/reference/UDS Core/IdAM/authentication-flows.md
UnicornChance: per offline conversation, agreed the doc was not serving its original intent. Moved extra info to a new dev docs directory that will hopefully be a dumping ground of useful developer documentation and usecases that we won't maintain in the code.

mjnagel (approved):

PR #299, opened 2025-01-09
feat: added trusted hosts necessary for istio ambient

sgettys:
## Description
adds trusted host for istio ambient with ztunnel ("*.pepr-uds-core-watcher.pepr-system.svc.cluster.local") and istio ambient with waypoint ("*.keycloak.svc.cluster.local")

Ambient traffic with only ztunnel will show as coming from the pepr pod directly since it only does L4 routing and connects directly to the Keycloak pod. Ambient traffic when Waypoint is deployed for Keycloak shows as originating from the Keycloak Waypoint proxy.

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


src/realm.json
UnicornChance, line 1: This looks good! Can we also add this to the [test realm.json](https://github.com/defenseunicorns/uds-identity-config/blob/main/src/test/cypress/realm.json) file ( for parity reasons ).

UnicornChance (approved):
LGTM

mjnagel (commented):
I think this makes sense and shouldn't hurt anything, but we may need to try and find a way to do this on an upgrade (the realm is only imported at install), potentially with Terraform + manual upgrade docs (two options). That could be a follow on PR if we want though.

UnicornChance:
That's a good call i didn't consider @mjnagel. [This](https://github.com/defenseunicorns/uds-identity-config/blob/main/docs/reference/UDS%20Core/IdAM/upgrading-versions.md) is where we've been documenting upgrade steps that require manual intervention, if we go that route.

slaskawi:
Added documentation

UnicornChance (approved):
LGTM

PR #336, opened 2025-02-11
feat: force re-login after changing credentials

### LINKED ISSUE #329: enable forced login upon password reset
### Is your feature request related to a problem? Please describe.
In [KC 26.1.1](https://github.com/defenseunicorns/uds-identity-config/pull/327) a new authenticator is enabled for forcing a user to login after a successful reset of their password.

At the moment, if a user resets their password they are automatically logged in. In a few usescases this presents a problem if that user shouldn't be using email/password credentials to login.

Currently this has been mitigated by disabling the ability to reset the password, but an additional defense in depth option would be beneficial here. This will probably be tied into the existing [authentication flow customization toggles](https://github.com/defenseunicorns/uds-identity-config/blob/main/docs/dev/authentication-flow-toggle-maps.md)

I believe at all times this should be enabled and should also be exposed ( undocumented to mitigate its overuse, since it's not recommended )  as an override.

slaskawi:
## Description

This Pull Request forces users to re-login after they change their credentials. 

More info: https://www.keycloak.org/docs/latest/release_notes/index.html#new-option-in-send-reset-email-to-force-a-login-after-reset-credentials

Since this is a low-priority hardening change, I skipped writing automated tests for it. 

## Related Issue

Fixes #329

## Type of change

## Checklist before merging

### DIFF
--- a/src/realm.json
+++ b/src/realm.json
@@ -2417,6 +2417,7 @@
                     "userSetupAllowed": false
                 },
                 {
+                    "authenticatorConfig": "reset-credentials-email",
                     "authenticator": "reset-credential-email",
                     "authenticatorFlow": false,
                     "requirement": "REQUIRED",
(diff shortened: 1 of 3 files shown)


src/realm.json
```diff
@@ -2896,6 +2897,13 @@
             "alias": "main",
             "config": {}
         },
+        {
```
slaskawi, line 2900: It seems I need to implement the same changes in `src/test/cypress/realm.json`

UnicornChance: Yeah, this looks good. Thanks for adding the upgrade docs! My only comment is adding this in the test realm as well ( like you said ).

slaskawi: Done!

PR #358, opened 2025-03-04
feat: pepr Keycloak Client management

slaskawi:
## Description

This Pull Request is the counterpart of https://github.com/defenseunicorns/uds-core/pull/1341 and introduces necessary plugins and Realm Configuration for using Client Credentials Grant by the UDS Operator to manage Keycloak Clients

The documentation for this feature will be delivered separately

## Related Issue

* Fixes https://github.com/defenseunicorns/uds-core/issues/1329
* Fixes https://github.com/defenseunicorns/uds-core/issues/1328
* Relates to https://github.com/defenseunicorns/uds-core/issues/1296
* [Technical Design](https://www.notion.so/Managing-Keycloak-Clients-from-Pepr-1a5e512f24fc80c4bc35d13b9d7a1367)

## Type of change

## Checklist before merging


.github/workflows/test.yaml
```diff
@@ -134,3 +134,25 @@ jobs:
 
       - name: Cypress Integration Tests
         run: uds run uds-core-integration-tests --no-progress
+
+  uds_core_cypress_integration_dynamic_client_registration:
```
slaskawi, line 138: Ask Keycloak Admins how to run this?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/ClientIdAndKubernetesSecretAuthenticator.java
```diff
@@ -0,0 +1,219 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import java.nio.file.Path;
+import java.util.*;
+
+import static java.nio.file.Files.readString;
+
+public class ClientIdAndKubernetesSecretAuthenticator extends AbstractClientAuthenticator {
```
slaskawi, line 33: public docs

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/ClientIdAndKubernetesSecretAuthenticator.java
```diff
@@ -0,0 +1,219 @@
...
+
+    public static final String PROVIDER_ID = "client-kubernetes-secret";
```
slaskawi, line 35: Can we find a better name?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -0,0 +1,85 @@
...
+import org.keycloak.services.clientpolicy.ClientPolicyException;
+import org.keycloak.services.clientpolicy.context.ClientCRUDContext;
+import org.keycloak.services.clientpolicy.executor.ClientPolicyExecutorProvider;
+import org.keycloak.services.clientpolicy.executor.FullScopeDisabledExecutorFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.invoke.MethodHandles;
+
+public class UDSClientPolicyPermissionsExecutor implements ClientPolicyExecutorProvider<ClientPolicyExecutorConfigurationRepresentation> {
```
slaskawi, line 22: Public docs

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public String getProviderId() {
+        return FullScopeDisabledExecutorFactory.PROVIDER_ID;
+    }
+
+    private void autoConfigure(ClientRepresentation rep) {
+        if (rep.getAttributes() == null)
+            rep.setAttributes(new java.util.HashMap<>());
+        rep.getAttributes().put(ATTRIBUTE_UDS_OPERATOR, ATTRIBUTE_UDS_OPERATOR_VALUE);
+        rep.setFullScopeAllowed(false);
```
slaskawi, line 82: Compare this with features we have in sso part of the Package spec

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutorFactory.java
```diff
@@ -0,0 +1,52 @@
...
+import org.keycloak.Config;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.KeycloakSessionFactory;
+import org.keycloak.provider.ProviderConfigProperty;
+import org.keycloak.services.clientpolicy.executor.ClientPolicyExecutorProviderFactory;
+
+import java.util.Collections;
+import java.util.List;
+
+public class UDSClientPolicyPermissionsExecutorFactory implements ClientPolicyExecutorProviderFactory {
```
slaskawi, line 17: Public docs

src/realm.json
```diff
@@ -802,6 +823,64 @@
                 "offline_access",
                 "microprofile-jwt"
             ]
+        },
+        {
+            "id": "16e3cc1f-89d1-4012-b40d-0bdecfe21d70",
+            "clientId": "uds-operator",
+            "name": "uds-operator",
+            "description": "A client used by the UDS Operator",
+            "rootUrl": "",
```
slaskawi, line 832: We don't need these default options

tasks.yaml
```diff
@@ -129,8 +135,9 @@ tasks:
   - name: clone-core
     description: "Clone UDS Core for integration testing"
     actions:
-      - cmd: rm -rf uds-core/
-      - cmd: git clone --branch ${CORE_VERSION} https://github.com/defenseunicorns/uds-core.git
+      - cmd: rm -rf uds-core
+#      - cmd: git clone --branch ${CORE_VERSION} https://github.com/defenseunicorns/uds-core.git
```
slaskawi, line 139: This Pull Request is in Draft as https://github.com/defenseunicorns/uds-core/pull/1341 needs to be merged first

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Override
+    public String getProviderId() {
+        return FullScopeDisabledExecutorFactory.PROVIDER_ID;
+    }
+
+    private void setUDSOperatorAsOwner(ClientRepresentation rep) {
+        if (rep.getAttributes() == null)
+            rep.setAttributes(new java.util.HashMap<>());
+        rep.getAttributes().put(ATTRIBUTE_UDS_OPERATOR, ATTRIBUTE_UDS_OPERATOR_VALUE);
```
slaskawi, line 102: We will probably need to mirror the Client Registration Policies here (potentially without a list of trusted hosts as this is already covered by the Authentication and Istio).

mjnagel: I think we probably would want to keep the restrictions but I might need to further review. If there are `Package` CRs that could be created, but fail reconciliation because of this policy - we may want to review our validator code and block `Package` CRs if the SSO config would cause an error (this would surface the failure quicker).

slaskawi: Ok, makes sense. I've implemented missing checks and enforcement.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/ClientIdAndKubernetesSecretAuthenticator.java
```diff
@@ -0,0 +1,222 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+    private String secretMountPath;
+
+    static String readMountedClientSecret(String secretMountPath, String clientId) throws IOException {
+        String mountedClientSecret = null;
+        Path mountedClientSecretPath = Path.of(secretMountPath).resolve(clientId);
+        logger.debug("Reading Client Secret from: {}", mountedClientSecretPath);
+
+        mountedClientSecret = readString(mountedClientSecretPath);
```
UnicornChance, line 55: is there any concern with whitespace here? wondering if it would be a good idea to add `.trim()` for the sake of safety.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/ClientIdAndKubernetesSecretAuthenticator.java
```diff
@@ -0,0 +1,222 @@
...
+
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import java.nio.file.Path;
+import java.util.*;
+
+import static java.nio.file.Files.readString;
+
+/**
+ * Keycloak Client Authenticator that uses client_id obtained from the request and client_secret obtained from a Kubernetes Secret mounted into the Keycloak Pod.
```
UnicornChance, line 34: Can we expand this comment to provide some more context of what's happening in this authenticator?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutorFactory.java
```diff
@@ -0,0 +1,55 @@
...
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.KeycloakSessionFactory;
+import org.keycloak.provider.ProviderConfigProperty;
+import org.keycloak.services.clientpolicy.executor.ClientPolicyExecutorProviderFactory;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Keycloak Client Policy Executor Factory for the UDS Operator.
```
UnicornChance, line 18: it wasnt immediately clear to me what the use of this was for and had to go dig through the keycloak docs to understand it a bit more. i think it would be good to add some more context here. 
```suggestion
 * Factory for creating and registering the UDS Client Policy Executor with Keycloak.
 *
 * This class is needed to integrate the custom UDS security hardening logic into Keycloak's
 * client management process. It creates instances of UDSClientPolicyPermissionsExecutor, which
 * enforce policies such as:
 * - Marking clients as created by the UDS Operator by setting the "created-by=uds-operator" attribute.
 * - Restricting operations on clients not owned by the UDS Operator or not following the naming convention.
 * - Disabling the "Full Scope Allowed" feature and limiting client scopes to an allowed set.
 *
 * Without this factory, Keycloak would not be able to instantiate and manage the custom policy
 * executor, and the UDS-specific client policies would not be enforced.
```

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/ClientIdAndKubernetesSecretAuthenticator.java
slaskawi: Of course! Added!

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -0,0 +1,153 @@
...
+ * * The Client can't use the Full Scope Allowed feature.
+ * * The Client can only use a limited set of Client Scopes (see {@code UDSClientPolicyPermissionsExecutor.ALLOWED_CLIENT_SCOPES}).
+ */
+public class UDSClientPolicyPermissionsExecutor implements ClientPolicyExecutorProvider<ClientPolicyExecutorConfigurationRepresentation> {
+
+    public static final String UDS_OPERATOR_CLIENT_ID = "uds-operator";
+    public static final String ATTRIBUTE_UDS_OPERATOR = "created-by";
+    public static final String ATTRIBUTE_UDS_OPERATOR_VALUE = "uds-operator";
+
+    public static final List<String> ALLOWED_CLIENT_SCOPES = List.of("openid",
```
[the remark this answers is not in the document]
slaskawi: I added a dev doc with proper comment. But I'm having second thoughts if we really should put any restriction here in place. After all, the Package CR already has `defaultClientScopes` as well as `protocolMappers`, so we expose ways to modify the Access Token.

On one had, adding custom Client Scopes gives Mission Heroes flexibility to include non-standard Claims in the tokens that might be necessary for their mission. When integrating a third part IdP (like Entra ID) with custom mappers, we need a custom claim to get it into the token.

On the other hand, from the security hardening perspective, there's no better way to prevent accidental misconfiguration and getting unwanted things into tokens as restricting Client Scopes (and transitively - Protocol Mappers as they are attached to Client Scopes). 

So we're looking at a balance between the flexibility and security here. **My personal opinion is to loosen up restrictions here and favor flexibility**.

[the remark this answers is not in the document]
slaskawi: What's your take on this @UnicornChance @mjnagel and @bburky ?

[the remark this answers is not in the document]
UnicornChance: I agree, with the package CR exposing the defualtClientScopes and protocolMappers, this will probably get confusing for end users. I don't feel strongly about this though, so if someone else thinks security is the better approach, i definitely see the value in that.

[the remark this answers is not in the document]
slaskawi: After discussing this in person, we decided to leave the list unrestricted now and harden this feature more in the future. This effort is tracked here: https://github.com/defenseunicorns/uds-identity-config/issues/385

The idea is to provide a configurable policy with sane defaults for both - allowed Protocol Mappers as well as Custom Claims.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -0,0 +1,153 @@
...
+            return null;
+        }
+        return clientCRUDContext.getAuthenticatedClient().getClientId();
+    }
+
+    boolean isOwnedByUDSOperator(ClientModel client) {
+        return ATTRIBUTE_UDS_OPERATOR_VALUE.equals(client.getAttribute(ATTRIBUTE_UDS_OPERATOR));
+    }
+
+    boolean hasBackwardsCompatibleClientName(ClientModel client) {
```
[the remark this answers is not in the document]
slaskawi: Note for reviewers: After doing a research, it turned out that there are too many different Client IDs out there to write a pattern to match them all. I decided to flip the logic and prevent the UDS Operator from accessing a few critical Clients from Keycloak and logging warnings if it tries to access some weird naming patterns. 

I think we can revisit this in a couple of releases and assume the Client Credentials has successfully been deployed and make this much more restrictive. 

This effort is tracked in https://github.com/defenseunicorns/uds-identity-config/issues/386

UnicornChance (dismissed):

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/ClientIdAndKubernetesSecretAuthenticator.java
```diff
@@ -0,0 +1,235 @@
...
+
+        if (!client.isEnabled()) {
+            context.failure(AuthenticationFlowError.CLIENT_DISABLED, null);
+            return;
+        }
+
+        if (client.isPublicClient()) {
+            context.success();
+            return;
+        }
```
mjnagel, line 143: Is this authenticator run for all clients, or just our `uds-operator` client? Assuming it's only for our operator client, would it be better to fail + throw some error here because the client _shouldn't_ be public? Okay moving forward with this as-is though - I think this only plays a role if an admin manually changed the client to set it to public.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -0,0 +1,132 @@
...
+            case "realm-management":
+            case "security-admin-console":
+            case "uds-operator":
+            logger.warn("Denying access to the Client with Client ID: {}. This Client is not allowed to be accessed by the UDS Operator.", client.getClientId());
+            return false;
+        }
+
+        if (!client.getClientId().startsWith("uds")) {
+            logger.debug("Allowing access to the Client with Client ID: {} in backwards compatibility mode. Please note, that this type of access will be denied in the future.", client.getClientId());
+        }
```
mjnagel, line 100: Okay leaving this as-is for now if you still see a use-case for it, but with the changes on the core side to add this attribute in dynamic mode I think we can avoid this backwards compatibility check?

mjnagel (commented):
LGTM overall - small comments, but the functionality seems to work as expected in testing.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
slaskawi: After thinking about this for a while, I think you're right. During the upgrade, Mission Heroes will first upgrade the UDS Operator, which will go through all the Packages and update the required Client Attribute. Then, Mission Heroes will start upgrading the Realm. 

I'll remove the backwards compatibility check but will also mention in the upgrade notes that if Mission Heroes hit any problems, it is advisable to disable the Policy and just it a bit more time to update everything.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/ClientIdAndKubernetesSecretAuthenticator.java
slaskawi: No, this needs to stay as is. The Client Authenticators are called in a chain (chain of responsibility pattern) - one after another with the order defined in the Client Credentials Authorization Flow. If we block here, we essentially block other Client Authenticators from evaluating. 

In theory we could make it a bit more restrictive by calling `context.attempted()` but I decided to make it more permissive. This way, you can use it along with other Authenticators as "required" giving less chances for user misconfiguration.

tasks.yaml
```diff
@@ -43,6 +47,8 @@ tasks:
       - task: dev-build
       - cmd: |
           k3d image import -c uds uds-core-config:keycloak
+          kubectl patch statefulset keycloak -n keycloak --type='merge' -p '{"metadata":{"labels":{"zarf.dev/agent":"ignore"}}}'
+          kubectl patch statefulset keycloak -n keycloak --type='merge' -p '{"spec":{"template":{"metadata":{"labels":{"zarf.dev/agent":"ignore"}}}}}'
```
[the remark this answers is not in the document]
slaskawi: If you don't mind, I'd like to leave it. Quite often I'm reusing my environment for working on UDS Core where I often deploy things using the bundle (at least I used to work with a few features I had to do this). In this case, I quickly ran into a situation where Zarf was interfering preventing the `dev-update-image`. With this change - it works in every case.

It's worth calling out that if the webhook is not there, this doesn't do anything.

mjnagel (approved):

tasks.yaml
UnicornChance: I've updated this to point at the main branch, will need to revert this back to latest core release after we release uds-core 0.39.0.

mjnagel (approved):

PR #360, opened 2025-03-04
fix: broken cypress tests

### LINKED ISSUE #348: local cypress e2e tests are failing + new sso.uds.dev test
### Steps to reproduce
1. Copying in the latest uds-core ( main branch ) and commenting out the clone uds-core task from `uds-core-integration-tests` task
2. running that task `uds run uds-core-integration-tests` 
3. I see failures on the `Existing Users` tests consistently 

### Additional test
Expand the verify logged in cypress command so that it makes sure the user profile page in the account portal is actually loading. There was [a bug](https://github.com/defenseunicorns/uds-identity-config/pull/347) that caused the account UI to return 200 status code but not actually load all the page content. Would be very beneficial to check this for the future.

### Expected result
CI works fine, but local does not. Need to investigate and understand why

UnicornChance:
## Description

It appears that in the latest version of cypress (14.1+) there was some optimizations made to the tests that has exposed a race condition that was undetected in the tests. ( could also be that the cypress tests are now synchronous by default whereas previous to the new cypress version they were asynchronous, however the release notes don't mention this).

The way that cypress manages injected CAC browser certs, means that it's there for all tests. Previously the registration test that actually uses that cert always got to it first, now the login user test gets there first and causes cypress to fail because it wasn't expecting to get a CAC pop up.

There is two avenues to fix this, i went with the 2nd option so that we could guarantee the outcome of tests now and in the future where there are additional tests added:

1. combine test into a single file, rather than three individual ones
     - simplest solution, only 9 tests total across the three files
     - does mean the tests need to be ordered correctly ( fragile in the long term )
     - can maintain the npm run command that gets all cypress tests regardless of name
     
2. add some logic so that the cac is only present in the browser for the registration tests.
     - this would allow us to keep individual files for the different test
     - also means the tests are truly individual from one another
     - requires that an ENV be passed in to the registration test

## Test locally:
`uds run uds-core-integration-tests` - run cypress tests against latest release of uds-core, should result in successful e2e tests.

## Related Issue

Fixes #348 

## Type of change

## Checklist before merging

### DIFF
--- a/src/test/cypress/package.json
+++ b/src/test/cypress/package.json
@@ -4,9 +4,9 @@
   "main": "index.js",
   "scripts": {
     "cy.open": "cypress open",
-    "cy.run": "cypress run",
+    "cy.run": "cross-env USE_CAC=true cypress run --spec 'e2e/registration.cy.ts' && cross-env USE_CAC=false cypress run --spec 'e2e/**/*.cy.ts,!e2e/registration.cy.ts'",
     "cy.run:registration": "cross-env USE_CAC=true cypress run --spec 'e2e/registration.cy.ts'",
-    "cy.run:others": "cross-env USE_CAC=false cypress run --spec 'e2e/login.cy.ts,e2e/group-authz.cy.ts'",
+    "cy.run:others": "cross-env USE_CAC=false cypress run --spec 'e2e/**/*.cy.ts,!e2e/registration.cy.ts'",
     "prettier": "prettier --write **/*.ts"
   },
   "devDependencies": {
--- a/src/test/cypress/package.json
+++ b/src/test/cypress/package.json
@@ -5,11 +5,14 @@
   "scripts": {
     "cy.open": "cypress open",
     "cy.run": "cypress run",
+    "cy.run:registration": "cross-env USE_CAC=true cypress run --spec 'e2e/registration.cy.ts'",
+    "cy.run:others": "cross-env USE_CAC=false cypress run --spec 'e2e/login.cy.ts,e2e/group-authz.cy.ts'",
     "prettier": "prettier --write **/*.ts"
   },
   "devDependencies": {
     "cypress": "^14.0.0",
     "prettier": "^3.2.5",
-    "typescript": "^5.4.3"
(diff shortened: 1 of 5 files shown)


src/test/cypress/package.json
```diff
@@ -5,11 +5,14 @@
   "scripts": {
     "cy.open": "cypress open",
     "cy.run": "cypress run",
```
mjnagel, line 7: Will this script fail as-is? If so, should we just remove it from the list here? Or alternatively could we run the other two commands here with `&&`?

src/test/cypress/package.json
```diff
@@ -5,11 +5,14 @@
...
+    "cy.run:registration": "cross-env USE_CAC=true cypress run --spec 'e2e/registration.cy.ts'",
+    "cy.run:others": "cross-env USE_CAC=false cypress run --spec 'e2e/login.cy.ts,e2e/group-authz.cy.ts'",
```
mjnagel, line 9: I haven't looked too closely at these test but are there any dependencies between them (i.e. does login require registration first)? If so - wonder if we can handle this cleanly...maybe related to the other comment and whether we could keep the `cy.run` task still.

src/test/cypress/package.json
UnicornChance: yeah thats a good call, the `cy.run` will fail as is.

src/test/cypress/package.json
UnicornChance: The tests are independent of each other, they can be run in any order or any subset. Previous to these changes that wasn't the case because of the CaC stuff.

slaskawi (dismissed):

src/test/cypress/package.json
UnicornChance: i've updated the scripts to have one that runs the registration tests and one that runs everything but the registrations tests. This is different from initial implementation because it no longer requires specifying each file that isn't registration tests.

Updated the cy.run to run those two scripts, and also updated the top level cypress tests task to use the cy.run again.

mjnagel (approved):
LGTM - great find/fix.

PR #378, opened 2025-03-18
fix!: limit aws groups mapper

### LINKED ISSUE #206: Limit Groups included in AWS Group Mapper
### Description

AWS has a character limit that is applied to the `Tag value`, which is where the groups are mapped from when utilizing the AWS Group Mapper. [This is a fixed limit on AWS's side.](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html)

This is a scaling issue when an organization has larger set of groups assigned to users. The current implementation of the AWS Group Mapper, creates a colon delimited list of **all** the groups of a user, regardless if they are needed or not.

@ntwkninja suggested a change of the implementation of the mapper to require the string `-aws-` in the keycloak group if its needed for AWS, and the mapper would only grab the groups with that string.

This would be a breaking change ( @ntwkninja  is the only user at this point so not a big problem ) and require some documentation, also might introduce duplicity of groups if a group is already defined without the aws string. Generally speaking though, doesn't seem too far fetched to have this requirement.

@rjferguson21 @bburky @mjnagel

UnicornChance:
## Description

BREAKING CHANGE: AWS has a character limit that is applied to the Tag value, which is where the groups are mapped from when utilizing the UDS Core AWS Group Mapper. This fix filters groups to only include groups containing a `-aws-` string.

[This is a fixed limit on AWS's side.](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html)

This is a scaling issue when an organization has larger set of groups assigned to users. The current implementation of the AWS Group Mapper, creates a colon delimited list of all the groups of a user, regardless if they are needed or not.

## Related Issue

Fixes #206 

## Type of change

## Checklist before merging


slaskawi (approved):

docs/reference/UDS Core/IdAM/plugin.md
```diff
@@ -13,7 +13,7 @@ A Keycloak plugin provides additional custom logic to our Keycloak deployment. B
...
+| [User AWS SAML Group Mapper](https://github.com/defenseunicorns/uds-identity-config/blob/v0.6.0/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/CustomAWSSAMLGroupMapper.java) | [SAML Mapper](https://www.keycloak.org/docs-api/latest/javadocs/org/keycloak/protocol/saml/mappers/AbstractSAMLProtocolMapper.html) | Amazon AppStream applications expect a specific group claim format when using Keycloak for authentication. This mapper filters the user’s groups to include only those with the substring `-aws-`. It then concatenates these qualified group paths into a colon-separated string that is passed in the SAML attribute. For example, if a user’s group hierarchy includes `/Core-aws/Admin-aws-test` and `/Core-aws/Auditor-aws-test`, the resulting SAML attribute value will be: `/parent-group/child-group1:/parent-group/child-group2`. |
```
mjnagel, line 16: Given that we use specific examples for the user's groups (`/Core-aws/Admin-aws-test` and `/Core-aws/Auditor-aws-test`) should we also use those same specific examples in the attribute value (i.e. `/Core-aws/Admin-aws-test:/Core-aws/Auditor-aws-test`)>

mjnagel (commented):
LGTM overall but let's make sure this is marked as a breaking change (`!` in the title) just in case this is being consumed somewhere we're not tracking.

mjnagel (approved):

PR #380, opened 2025-03-19
feat: webauthn and mfa everywhere

### LINKED ISSUE #236: Add Biometric / WebAuthn Enrollment and Authentication

- **Given** a user is setting up multi-factor authentication in Keycloak
- **When** the user reaches the MFA setup stage in the registration and authentication flows
- **Then** WebAuthn Passwordless Authentication (including biometric readers like TouchID, FaceID, Fingerprints, etc.) can be used for MFA

### Describe things you've considered
- The **_initial login_**/self-registration flows would need to be modified to include WebAuthn / Passwordless _**enrollment**_ (registration flow)
- Subsequent **_authentication_** flows (i.e. navigating to a mission app within UDS) would need to be updated to support WebAuthn / Biometric authentication (browser based authentication)

### Additional context
- [Medium Article](https://embesozzi.medium.com/keycloak-workshop-for-step-up-with-mfa-biometrics-authentication-passkeys-b7020ea9ae1b)
- [Keycloak Documentation](https://www.keycloak.org/docs/latest/server_admin/#webauthn_server_administration_guide)

UnicornChance:
## Description
This PR adds support for configuring WebAuthn MFA for x509 (CAC) users. Historically, our implementation assumed that when a user authenticated with x509, no further authentication (e.g., OTP or passkey) was necessary. However, feedback from certain environments indicated a requirement for MFA even for x509 users.

## Key Changes
- Conditional MFA for x509 Users:
We’ve introduced conditional logic in our custom registration validation plugin. This logic now allows administrators to optionally require MFA (via OTP and/or WebAuthn Register Passwordless) for x509 users. The new behavior is controlled via the environment variable X509_MFA_ENABLED. When enabled (set to true), the plugin adds the required actions for OTP and WebAuthn even for x509 users; otherwise, it applies these only to local (username/password) users.

- Environment Variable Mapping:
The new configuration settings are mapped in the same manner as our existing realm and theme environment variables. For example, the following variables in our values.yaml/uds-bundle.yaml are now available:

  - X509_MFA_ENABLED: Controls whether x509 users must complete additional MFA.
  - OTP_ENABLED, WEBAUTHN_ENABLED, and the derived MFA_FLOW_ENABLED remain in place to allow flexible MFA configuration.

- Backward Compatibility:
By default, the new behavior is disabled (i.e. X509_MFA_ENABLED=false) to ensure that existing deployments do not break. Administrators can enable it as needed for environments with stricter security requirements.

**[uds-core PR that needs to merge first](https://github.com/defenseunicorns/uds-core/pull/1370)**

## Testing
1. To setup our testing we need to do some manual changes first, comment out this line in the tasks file: https://github.com/defenseunicorns/uds-identity-config/blob/main/tasks.yaml#L111
2. then in the uds-identity-config directory we need to clone in uds-core and then checkout the correct branch with the changes in uds-core for this PR: `git clone https://github.com/defenseunicorns/uds-core.git && cd uds-core && git checkout webauthn-mfa && cd ../`
3. now we should have the default use-case of this new implementation and we can run `uds run uds-core-integration-tests` to get everything running
4. once everything has come up, verify that everything looks/feels/works as it used to ( there should be no changes by default )
5. Now to test the new variables and webauth flows change the [new variables](https://github.com/defenseunicorns/uds-core/blob/webauthn-mfa/src/keycloak/chart/values.yaml#L49-L55) in the [bundle](https://github.com/defenseunicorns/uds-identity-config/blob/webauthn-mfa/bundles/uds-bundle.yaml#L76-L81) in different combinations and see if anything breaks. For detailed docs on different configurations [see this PR's dev/doc update](https://github.com/defenseunicorns/uds-identity-config/blob/webauthn-mfa/docs/dev/authentication-flow-toggle-maps.md) - see the common configurations section at the very bottom.
6. after changes to bundle build and deploy again ( may be beneficial to comment out the [cypress tests from the uds-core-integration-tests task](https://github.com/defenseunicorns/uds-identity-config/blob/webauthn-mfa/tasks.yaml#L114) as well as the [monitoring bundle definition](https://github.com/defenseunicorns/uds-identity-config/blob/webauthn-mfa/bundles/uds-bundle.yaml#L83-L86) to speed things up)
7. There shouldn't be any admin portal configurations, so accessing `sso.uds.dev` should present any changes made.

## Related Issue

Fixes #236 

## Type of change

## Checklist before merging


src/theme/login/webauthn-authenticate.ftl
```diff
@@ -0,0 +1,118 @@
+<#import "template.ftl" as layout>
```
mjnagel, line 1: Not super sure where this comment belongs but I think it would be helpful to update the theming/buttons around the webauthn during registration. Currently it's not super clear what this register button is for (and I'm not sure if the sign out button should be there too?).

docs/reference/UDS Core/IdAM/authentication-flows.md
```diff
@@ -30,6 +30,36 @@ UDS Core comes equipped with a robust authentication framework that supports mul
 
 ![Authentication Flow Options](https://github.com/defenseunicorns/uds-identity-config/blob/main/docs/.images/diagrams/uds-core-auth-flows-options.svg?raw=true)
```
mjnagel, line 31: Do we also need an update to the diagram here? Currently it doesn't show x509 supporting MFA (also don't know if we'd want to show both OTP and WebAuthn).

src/theme/login/webauthn-authenticate.ftl
mjnagel, line 1: Also _potentially_ theme related - on the my account pages we should make sure we have access to view, edit, delete webauthn MFA. Currently I only see the password and otp 2fa sections.

src/realm.json
```diff
@@ -2157,10 +2156,18 @@
...
-                    "requirement": "REQUIRED",
+                    "requirement": "${OTP_FLOW_ENABLED:REQUIRED}",
                     "priority": 1,
                     "autheticatorFlow": false,
                     "userSetupAllowed": false
+                },
+                {
+                    "authenticator": "webauthn-authenticator-passwordless",
+                    "authenticatorFlow": false,
+                    "requirement": "${WEBAUTHN_FLOW_ENABLED:REQUIRED}",
```
slaskawi, line 2167: Why do we set it to required instead of alternative? If that's because it's the last in the Authentication Flow, what shall we do with "Condition - user configured"? Should it also be required?

slaskawi (dismissed):
The PR LGTM with one small question inlined in the text. 

I managed to get this working on MacOS using the Touch ID. Seems like a great MFA approach!

obraz

mjnagel (commented):
One more general comment just to make sure it's present here somewhere: This currently only adds webauthn support as an additional authentication factor. Users will need to have a separate "primary" auth factor (user/pass or x509) as well. While we could add support for "webauthn only" auth, we don't currently see the need for that - this could be reevaluated and supported in the need arose though.

src/realm.json
UnicornChance: It we set both OTP and WebAuthn to alternative, that would mean we only want one of those options. My intent with required was that we could control wether those 2FA options are required or not. Based on what was asked from Delivery, they would both be required in the future. However you make a good point that if someone wanted either or, we wouldn't have a configuration for that. I believe the user-configured is required all the time. Not sure if im answering your question here.

src/theme/login/webauthn-authenticate.ftl
UnicornChance: Agreed, this required a rework. I didn't realize i had been using the wrong `WebAuthn` Authenticator for this. So instead of using the `WebAuthn Register Passwordless` we're now using the `WebAuthn Register` required action. This sets up the 2FA the way keycloak intended, resulting in a viewable Passkey entry in the account console. Interestingly it seems that keycloak doesn't allow for deleting the old passkey, but allows for adding a new passkey and keeps track of both.
account-signing-in

double-passkey-auth

docs/reference/UDS Core/IdAM/authentication-flows.md
UnicornChance: updated this diagram

src/realm.json
slaskawi: aaaah got it! Thanks for explaining.

slaskawi (dismissed):

docs/reference/UDS Core/IdAM/upgrading-versions.md
```diff
@@ -10,65 +10,99 @@ This doc contains important information for upgrading uds-identity-config versio
...
+      - Check UDS Operator logs and verify if there are no errors
+         - Use `uds zarf tools kubectl logs deploy/pepr-uds-core-watcher -n pepr-system | grep "Client Credentials Keycloak Client is available"` command to verify if the UDS Operator uses the Client Credentials flow.
 
 After introducing the above changes, please ensure all Packages are reconciled correctly and there are no errors. If for some reason you see the UDS Operator throwing errors with `The Client doesn't have the created-by=uds-operator attribute. Rejecting request`, you need to disable the `UDS Client Policy` and give it a bit more time to process all the Packages.
 
+---
+
+In uds-identity-config version 0.11.0 we incorporated some big changes around MFA.
+- Previous versions didn't allow for MFA on the X509 Authentication flow. Now that can be configured to required additional factors of authentication. By default this is disabled and will need to be enabled.
+- Additionally, we've added support of WebAuthn MFA. This can assume many different form such as biometrics, passkeys, etc. This also is disabled by default and is only used as an MFA option.
```
mjnagel, line 71:
```suggestion
- Additionally, we've added support of WebAuthn MFA. This can assume many different forms such as biometrics, passkeys, etc. This also is disabled by default and is only used as an MFA option.
```

docs/reference/UDS Core/IdAM/upgrading-versions.md
```diff
@@ -10,65 +10,99 @@ This doc contains important information for upgrading uds-identity-config versio
...
+
+If wanting configure the MFA everywhere with both OTP and WebAuthn options, the following steps will help to manually configure these options:
```
mjnagel, line 73:
```suggestion
If wanting to configure the MFA everywhere with both OTP and WebAuthn options, the following steps will help to manually configure these options on an upgrade:
```

src/theme/login/webauthn-authenticate.ftl
mjnagel: LGTM!

mjnagel: I'm not sure if I'm doing something wrong but I'm still not seeing options in the account console for passkeys. For my overrides right now I have:
```yaml
            - path: realmAuthFlows
              value:
                USERNAME_PASSWORD_AUTH_ENABLED: true
                X509_AUTH_ENABLED: true
                SOCIAL_AUTH_ENABLED: true
                OTP_ENABLED: false
                WEBAUTHN_ENABLED: true
```
I also tested with OTP enabled but that didn't seem to make a difference.

mjnagel: I'm also a bit confused/concerned by there being no way to delete webauthn creds - on the admin side I'm able to delete a user's passkey (and did validate that the user is prompted to add a new one on next login). I think for OTP the account console provides options to add and remove so it's interesting that this doesn't seem to be the case for webauthn.

mjnagel (approved):

src/theme/login/webauthn-authenticate.ftl
mjnagel: Theme and ability to add new passkeys all works great with the latest commits/fixes. On the "deletion of passkey" we determined it's not required for this first iteration since that can still be done via the admin console. We should re-evaluate the feasibility of this in the future though.

PR #398, opened 2025-03-28
fix: add credential deletion configuration

### LINKED ISSUE #397: Allow deleting a passkey
## Description

Allow deleting a users passkey so that they can reset that passkey if necessary.

UnicornChance:
## Description
With the addition of WebAuthn Passkeys, there wasn't a way for a user to delete and re-create a passkey attached to their accound. This change allows for that use case.

## Testing Steps
1. Clone in the uds-core branch `git clone https://github.com/defenseunicorns/uds-core.git && cd uds-core && git checkout allow-deleting-a-passkey && cd ../`
2. comment out the task `clone-core` here: https://github.com/defenseunicorns/uds-identity-config/blob/main/tasks.yaml#L117
3. deploy identity config `uds run uds-core-integration-tests`
4. navigate to `sso.uds.dev` and register a new user with a passkey
5. in the users account console go to the `Account Security` -> `Signing in` page
6. in the `Two Factor Authentication` section you should see an entry for the passkeys, and notably beside those there should **not** be a bold red `Delete` button
7. configure the bundle and disable the registration fields:
```yaml
            - path: realmAuthFlows
              value:
                USERNAME_PASSWORD_AUTH_ENABLED: true
                X509_AUTH_ENABLED: true
                SOCIAL_AUTH_ENABLED: true
                OTP_ENABLED: false
                WEBAUTHN_ENABLED: true
                X509_MFA_ENABLED: true
```
8. deploy identity config again `uds run uds-core-integration-tests`
9. This time register a user with a passkey and verify that there is a big red `Delete` button next to the passkey

## Related Issue

Fixes #397
Relates to this UDS Core PR

## Type of change

## Checklist before merging


src/realm.json
```diff
@@ -3194,6 +3194,15 @@
             "defaultAction": false,
             "priority": 1003,
             "config": {}
+        },
+        {
```
mjnagel, line 3198: Does this need another upgrade note to enable this?

slaskawi: I absolutely agree with @mjnagel here - we should add a new entry in the upgrade guide

UnicornChance: Udpated the upgrade doc, good call. really small change since we already have webauthn section and this is related.

mjnagel (approved):

PR #400, opened 2025-03-31
feat: configurable Terms and Conditions

### LINKED ISSUE #368: Configurable Terms and Conditions
### Is your feature request related to a problem? Please describe.
Currently when signing into keycloak for the first time, you have to accept the Terms and Conditions (T&C) regardless of the service you are accessing. After that whenever you open new services or applications (opening `gitlab.mydomain.com` when you are already signed into `confluence.mydomain.com`), you have to once again accept the T&C. I want the T&C banner to be configurable so it only shows on initial login.

- **Given** A user is signed in and not timed out
- **When** I access another service behind the same keycloak
- **Then** I am not asked to accept the T&C again

slaskawi:
## Description

This Pull Request introduces configurable Terms and Conditions.

Leaving this plugin unconfigured or setting the "UDS Operator Group Authentication Validation config" explicitly and marking "Display Terms and Conditions only per user session" to true will prevent Terms and Conditions from displaying if a user already accepted it within his session. Here's an example:

1. Log into Grafana with your credentials
2. Accept Terms and Conditions
3. Log into Neuvector with your credentials
4. Note that Terms and Conditions are not displayed!

When the "Display Terms and Conditions only per user session" is configured to false, it falls back into the previous behavior. 

obraz

## Related Issue

Fixes #368

## How to test this change

1. Run `uds run uds-core-integration-tests`
2. Run `cd uds-core && uds run create:single-layer-callable --set LAYER=runtime-security && uds zarf package deploy build/zarf-package-core-runtime-security-arm64-0.38.0.tar.zst`
3. Go to Keycloak Admin console and flip the Required Actions - Terms and Conditions on.

!! This test will reuse the Integration tests and a custom theme (which is not pretty - Sorry!) !!

5. Go to https://grafana.admin.uds.dev/
5a. Log in
5b. Accept Terms and Conditions
6. Go to https://neuvector.admin.uds.dev/
6a. Log in
6b. Note, there's no Terms and Conditions displayed
6c. Verify the logs with `kubectl logs keycloak-0 -n keycloak -f | grep "RequireGroupAuthenticator"` and ensure you see `User already accepted Terms and Conditions for this session. Skipping...`

## Type of change

## Checklist before merging


src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -121,11 +119,49 @@ private String buildGroupPath(GroupModel group) {
...
     }
 
-    private void success(final AuthenticationFlowContext context, final UserModel user) {
-        user.addRequiredAction("TERMS_AND_CONDITIONS");
+    protected void success(final AuthenticationFlowContext context, final UserModel user) {
+        boolean shouldAddTAC = true;
+
+        if (isConditionalTACActive(context)) {
+            String parentSessionId = context.getAuthenticationSession().getParentSession().getId();
+            LOGGER.debugf("Conditional TAC, parent session ID: %s", parentSessionId);
```
UnicornChance, line 127: These two debug logs expose session id's, which i think could be used for session hijacking if logs were ever exposed. Also could be tied back to users which would mean they are sensitive.

Couple ideas would be just removing the log or could truncate the session id.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -121,11 +119,49 @@ private String buildGroupPath(GroupModel group) {
...
+            if (user.getAttributes().get(TAC_USER_ATTRIBUTE) != null) {
+                String userSessionId = user.getAttributes().get(TAC_USER_ATTRIBUTE).get(0);
+                LOGGER.debugf("Conditional TAC, user session ID: %s", parentSessionId);
+                if (parentSessionId.equals(userSessionId)) {
+                    shouldAddTAC = false;
+                    LOGGER.info("User already accepted Terms and Conditions for this session. Skipping...");
```
UnicornChance, line 133: Not something i feel strongly about but in my opinion these logs should be debug level, trying to avoid adding more noise to the already overwhelming keycloak logs.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
```diff
@@ -121,11 +119,49 @@ private String buildGroupPath(GroupModel group) {
...
+                } else {
+                    LOGGER.info("Stale session detected, asking user to accept Terms and Conditions again");
+                }
+            } else {
+                LOGGER.info("User didn't accept Terms and Conditions. Requesting accepting it");
```
UnicornChance, line 138:
```suggestion
                LOGGER.info("User hasn't accepted Terms and Conditions. Requesting user accept them.");
```

nit: i think the wording is a little confusing

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
slaskawi: This is a very, very good catch! 

I checked the Keycloak codebase and indeed, the Root Session ID is taken from the cookie: https://github.com/keycloak/keycloak/blob/594218382d293c0adc20f80d6302e5c491ca2191/services/src/main/java/org/keycloak/protocol/AuthorizationEndpointBase.java#L184-L194C93

I'll remove this and convert all the logger calls to trace.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/RequireGroupAuthenticator.java
slaskawi: Makes sense! I changed the logging level to trace so that the TAC functionality is silent in runtime.

UnicornChance (commented):
I'm probably not configuring something or misunderstanding something but i was unable to replicate this behavior with the given testing steps.

slaskawi:
@UnicornChance Updated the verification steps. Please feel free to have a look at it whenever you find a moment.

UnicornChance:
As discussed synchronously, i think my original issue was cookie problem. Was able to verify everything worked this time around! LGTM!

UnicornChance (approved):

Issue #385, opened 2025-03-25
Enhance the uds-operator-permissions Policy Executor to assign allowed Protocol Mapper and Custom Claims


slaskawi:
### Is your feature request related to a problem? Please describe.

When working on https://github.com/defenseunicorns/uds-identity-config/pull/358, we realized that we need a way to dynamically choose what Protocol Mappers we allow to be set on the Client as well as what Custom Claims we allow. This configuration needs to happen dynamically with the help of UI. The UDS Platform will provide a sane set of defaults but also will enable a way to customize them.

PR #428, opened 2025-04-14
feat: uds policy hardening

slaskawi:
## Description

This Pull Request introduces new configuration options for the `UDSClientPolicyPermissionsExecutor` that limit what the UDS Operator can do in Keycloak. 

## Related Issue

Fixes: https://github.com/defenseunicorns/uds-identity-config/issues/385

## Type of change

## Checklist before merging

### DIFF
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
@@ -5,9 +5,18 @@
 
 package com.defenseunicorns.uds.keycloak.plugin.clientpolicy.executor;
 
+import com.defenseunicorns.uds.keycloak.plugin.CustomAWSSAMLGroupMapper;
+import com.defenseunicorns.uds.keycloak.plugin.CustomGroupPathMapper;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.keycloak.events.Errors;
 import org.keycloak.models.ClientModel;
-import org.keycloak.representations.idm.ClientPolicyExecutorConfigurationRepresentation;
+import org.keycloak.models.ClientScopeModel;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.protocol.ProtocolMapper;
+import org.keycloak.protocol.oidc.mappers.*;
+import org.keycloak.protocol.saml.mappers.RoleListMapper;
+import org.keycloak.protocol.saml.mappers.UserAttributeStatementMapper;
+import org.keycloak.protocol.saml.mappers.UserPropertyAttributeStatementMapper;
 import org.keycloak.representations.idm.ClientRepresentation;
 import org.keycloak.services.clientpolicy.ClientPolicyContext;
 import org.keycloak.services.clientpolicy.ClientPolicyException;
@@ -18,6 +27,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.lang.invoke.MethodHandles;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * Keycloak Client Policy Executor for the UDS Operator.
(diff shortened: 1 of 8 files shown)


src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -82,13 +137,58 @@ boolean isOwnedByUDSOperator(ClientModel client) {
...
+        if (rep.getProtocolMappers() != null && !rep.getProtocolMappers().isEmpty()) {
+            rep.getProtocolMappers()
+                    .removeIf(mapper -> !configuration.getAllowedProtocolMappers().contains(mapper.getProtocolMapper()));
+        }
+    }
+
+    void setAllowedCustomClientScopes(ClientRepresentation rep) {
+        if (rep.getDefaultClientScopes() != null && !rep.getDefaultClientScopes().isEmpty()) {
+            rep.getDefaultClientScopes()
+                    .removeIf(scope -> !configuration.getAllowedClientScopes().contains(scope));
```
slaskawi, line 150: We probably need an extra check for `configuration.getAllowedClientScopes()` to keep operating in case of misconfiguration or Keycloak SPI changes. I'll add it along with other comments.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutorFactory.java
```diff
@@ -60,6 +115,9 @@ public String getHelpText() {
 
     @Override
     public List<ProviderConfigProperty> getConfigProperties() {
-        return Collections.emptyList();
+        return List.of(additionalAllowedProtocolMappers,
```
UnicornChance, line 118: I think it would be safer to use `new ArrayList<>(...)` in case the list is mutated downstream someday

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutorFactory.java
```diff
@@ -8,40 +8,95 @@
...
+        useDefaultAllowedProtocolMappers.setType(ProviderConfigProperty.BOOLEAN_TYPE);
+        useDefaultAllowedProtocolMappers.setDefaultValue(Boolean.TRUE.toString());
+        configProperties.add(useDefaultAllowedProtocolMappers);
+
+        additionalAllowedClientScopes.setName(ADDITIONAL_ALLOWED_CLIENT_SCOPES);
+        additionalAllowedClientScopes.setLabel("Additional Allowed Client Scopes");
+        additionalAllowedClientScopes.setHelpText("Additional Allowed Client Scopes");
+        additionalAllowedClientScopes.setType(ProviderConfigProperty.MULTIVALUED_STRING_TYPE);
+        // Unfortunately, when we're creating a Client Policy we don't have access to Client Scopes yet.
+        // This is different in
```
UnicornChance, line 68: nit: not sure i understand this comment, maybe it was cut short or something?

src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutorTest.java
```diff
@@ -74,4 +104,79 @@ public void shouldEnforceFullScopeDisabled() {
...
+        }};
+
+        ClientRepresentation rep = new ClientRepresentation();
+        // We need a modifiable List implementation here - that's why it's wrapped with an ArrayList
+        rep.setProtocolMappers(new ArrayList<>(List.of(allowedProtocolMapper, disallowedProtocolMapper)));
+
+        // when
+        executor.enforceClientSettings(rep);
+
+        // then
```
UnicornChance, line 125: this is totally personal preference here, so disregard if you prefer otherwise. I like to add an explicit negative case test for the invalid protocol mapper, asserting its not present.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
slaskawi: Done

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutorFactory.java
slaskawi: This is actually fine here!

The `List.of` returns an immutable collection for showing the UI. This collection should never be changed by none of the components.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutorFactory.java
slaskawi: This is a leftover, thanks for pointing it out. 

I was experiment if we can pre-populate the list of custom claims. Turns out that without changes to Keycloak SPI, we can't.

src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutorTest.java
slaskawi: This is actually what's happening in this test. 

In the line 200, we start with 2 Protocol Mappers added to the Client. One of them is valid and one is invalid: 

```java
        rep.setProtocolMappers(new ArrayList<>(List.of(allowedProtocolMapper, disallowedProtocolMapper)));
```

Then, we enforce the settings that should remove the invalid Protocol Mapper. We then check the asserts:

```java
        assertEquals(1, rep.getProtocolMappers().size());
        assertEquals(allowedProtocolMapper.getId(), rep.getProtocolMappers().get(0).getId());
```

Out of the 2 Protocol Mappers we started with we have only 1. The next assert checks if the one that's in there is the valid one. If both asserts pass, we know that the invalid has been removed. 

Perhaps you'd like me to rewrite this in some specific way? Would you mind to give me some more suggestions?

UnicornChance: Yeah agreed, your test is totally fine as is. It's definitely my personal preference to use something like:
```
assertFalse(rep.getProtocolMappers().stream()
    .anyMatch(mapper -> "invalid".equals(mapper.getProtocolMapper())));
```
which tests explicitly that the invalid mapper is not present, this can be assumed based on your test though so happy to leave it as is.

slaskawi: I'll add it - no problem!

UnicornChance (approved):
LGTM, good work!

Issue #36, opened 2024-03-20
FIPS Support


UnicornChance:
#### At this time this is more of a discussion placeholder than an actual intent to implement.

FIPS is not supported at the moment. IF we want to move to supporting FIPS in our identity config we will need to spend some time understanding the different bouncy-castle libraries. This primarily impacts the [x509 plugin source code](https://github.com/defenseunicorns/uds-identity-config/tree/main/src/plugin), but will probably affect the other pieces as well.

At the moment we utilize the `bcpg-fips`  library, but based on bouncy-castle documentation that also requires at least a library for FIPS algorithms and potentially also using the `bc-fips` general FIPS library that isn't specific to OpenPGP.

Current bouncy-castle and Keycloak crypto libraries:
```      
        <dependency>
            <groupId>org.keycloak</groupId>
            <artifactId>keycloak-crypto-default</artifactId>
            <version>${keycloak.version}</version>
        </dependency>
        <dependency>
            <groupId>org.bouncycastle</groupId>
            <artifactId>bcpg-fips</artifactId>
            <version>1.0.7.1</version>
        </dependency>
```

Libraries that should be looked into:
```
        <dependency>
            <groupId>org.keycloak</groupId>
            <artifactId>keycloak-crypto-fips1402</artifactId>
            <version>${keycloak.version}</version>
        </dependency>
        <dependency>
            <groupId>org.bouncycastle</groupId>
            <artifactId>bc-fips</artifactId>
            <version>1.0.2.4</version>
        </dependency> 

```

### Describe the solution you'd like
- All testing should pass (integration tests, plugin unit tests, etc). 
- The provided identity config should be FIPS compliant, regardless of the surrounding environment.

### Testing Gotcha's
- While testing changes to libraries be aware that sometimes the order of the libraries in the pom.xml can cause errors. @rjferguson21 and @UnicornChance  experienced this behavior when importing the `keycloak-crypto-defualt` library after the `bcpg-fips` library. 

### Links
[Official Keycloak FIPS140-2 Docs](https://www.keycloak.org/server/fips)
[Maven Repo description of bcpg-fips](https://mvnrepository.com/artifact/org.bouncycastle/bcpg-fips)

PR #434, opened 2025-04-16
feat: fips mode

slaskawi:
## Description

This Pull Request introduces FIPS mode to Keycloak.

The work here depends on https://github.com/defenseunicorns/uds-core/pull/1469

⚠️ The work in this Pull Request uses the v0.39.0 tag in UDS Core. Effectively this means that it won't use turn FIPS on as https://github.com/defenseunicorns/uds-core/pull/1469 is missing. The tests for this PR have been executed here: https://github.com/defenseunicorns/uds-identity-config/pull/446

## Related Issue

Relates to https://github.com/defenseunicorns/uds-identity-config/issues/36

## Type of change

## Checklist before merging

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -16,11 +16,10 @@ COPY plugin/src ./src
 
 RUN mvn clean package
 
-RUN mvn dependency:copy-dependencies -DoutputDirectory=target/lib
-
 # https://www.keycloak.org/server/fips#_bouncycastle_fips_bits
 # Keycloak FIPS mode requires additional libraries. This way we ensure its version is exactly
 # the same as use used by Keycloak Server.
+RUN mvn dependency:copy-dependencies -DoutputDirectory=target/lib
 RUN mkdir -p target/fips/libs
 RUN cp target/lib/bc-fips-*.jar target/fips/libs
 RUN cp target/lib/bctls-fips-*.jar target/fips/libs
@@ -68,7 +67,8 @@ COPY --chown=nonroot --from=plugin /home/build/target/*.jar .
 # FIPS dependencies
 RUN mkdir -p ./fips/libs
 RUN chown -R nonroot:nonroot ./fips/libs
-COPY --chown=nonroot --from=plugin /home/build/target/fips/libs/*.jar ./fips/libs
+COPY --chown=nonroot --from=plugin /home/build/target/fips ./
+
 
 # Copy the DOD Unclass PEM and truststore for Istio & Keycloak Auth
 COPY --chown=nonroot --from=truststore /home/build/authorized_certs.pem .
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -11,9 +11,22 @@ FROM cgr.dev/chainguard/maven:openjdk-17 AS plugin
 WORKDIR /home/build
 
 COPY plugin/pom.xml .
+RUN mvn dependency:go-offline
 COPY plugin/src ./src
 
 RUN mvn clean package
+
+RUN mvn dependency:copy-dependencies -DoutputDirectory=target/lib
+# https://www.keycloak.org/server/fips#_bouncycastle_fips_bits
+RUN mkdir -p target/fips/libs
+RUN cp target/lib/bc-fips-*.jar target/fips/libs
+RUN cp target/lib/bctls-fips-*.jar target/fips/libs
+RUN cp target/lib/bcpkix-fips-*.jar target/fips/libs
+RUN cp target/lib/bcutil-fips-*.jar target/fips/libs
+
+RUN pwd
+RUN ls -lR target/fips/libs
+RUN ls /home/build/target/fips/libs/*.jar
 
 # Include anything in the extra-jars directory in the target directory
 # We will filter out in the next build stage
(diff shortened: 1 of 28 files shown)


src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/UpdateX509Test.java
```diff
@@ -122,7 +123,8 @@ public void setupMockBehavior() throws Exception {
                     return Stream.of(userModel);
                 });
 
-        CryptoIntegration.init(this.getClass().getClassLoader());
+//        CryptoIntegration.init(this.getClass().getClassLoader());
+//        CryptoIntegration.setProvider(new FIPS1402Provider());
```
slaskawi, line 127: We can't reinitialize the provider as the Java Security Provider substitution happens at the JVM level. PowerMock uses separated Class Loaders to do mocks. This obviously causes errors (as the Security Providers are loaded by different Class Loaders).

src/Dockerfile
```diff
@@ -11,10 +11,22 @@ FROM cgr.dev/chainguard/maven:openjdk-17 AS plugin
 WORKDIR /home/build
 
 COPY plugin/pom.xml .
+RUN mvn dependency:go-offline
 COPY plugin/src ./src
 
 RUN mvn clean package
 
+RUN mvn dependency:copy-dependencies -DoutputDirectory=target/lib
```
slaskawi, line 19: Move this down

src/Dockerfile
```diff
@@ -53,6 +65,11 @@ WORKDIR /home/nonroot
 # Copy Jars from the plugin build stage
 COPY --chown=nonroot --from=plugin /home/build/target/*.jar .
 
+# FIPS dependencies
+RUN mkdir -p ./fips/libs
+RUN chown -R nonroot:nonroot ./fips/libs
```
slaskawi, line 70: Check if we can simplify this?

src/test/cypress/e2e/registration.cy.ts
```diff
@@ -33,7 +33,7 @@ describe("CAC Registration Flow", () => {
     cy.get(".form-group .alert-info").should("be.visible").contains("h2", "DoD PKI Detected");
     cy.get(".form-group #certificate_subjectDN")
       .should("be.visible")
-      .contains("C=US,ST=Colorado,L=Colorado Springs,O=Defense Unicorns,CN=uds.dev");
+      .contains("C=US").contains("ST=Colorado").contains("L=Colorado Springs").contains("O=Defense Unicorns").contains("CN=uds.dev")
```
slaskawi, line 36: Add explanation

UnicornChance (commented):
A bit vague but i think documenting how our FIPS flow works would be useful, especially as we move into STIG and compliance documentation things. Also would be beneficial for Growth i would think if we can directly point them to a doc when they have FIPS questions.

src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/RegistrationValidation2Test.java
```diff
@@ -121,8 +121,6 @@ public void setupMockBehavior() throws Exception {
                 .thenAnswer((stream) -> {
                     return Stream.of(userModel);
                 });
-
-        CryptoIntegration.init(this.getClass().getClassLoader());
```
UnicornChance, line 125: Can also remove the CryptoIntegration import

src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/RegistrationX509PasswordTest.java
```diff
@@ -5,6 +5,7 @@
 
 package com.defenseunicorns.uds.keycloak.plugin;
 
+import org.keycloak.crypto.fips.FIPS1402Provider;
```
UnicornChance, line 8: i might be missing it but is this import actually used?

src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/RegistrationX509PasswordTest.java
```diff
@@ -127,8 +128,6 @@ public void setupMockBehavior() throws Exception {
                 .thenAnswer((stream) -> {
                     return Stream.of(userModel);
                 });
-
-        CryptoIntegration.init(this.getClass().getClassLoader());
```
UnicornChance, line 131: Can also remove the CryptoIntegration import

UnicornChance: same comment for the UpdateX509Test.java and RegistrationValidation2Test.java files

tasks.yaml
```diff
@@ -41,7 +41,7 @@ tasks:
   - name: dev-build
     description: "Build the image locally for dev"
     actions:
-      - cmd: docker build src -t uds-core-config:keycloak --no-cache
+      - cmd: docker build src -t uds-core-config:keycloak
```
UnicornChance, line 44: could removing the cache cause any weird issues between builds? I'm assuming this is being removed to cut down on image pull time.

src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/RegistrationX509PasswordTest.java
slaskawi: Optimized imports in all touched files.

tasks.yaml
slaskawi: Yes! 

I try to be careful when introducing this but the biggest improvement comes with downloading the Maven dependencies. In the modified Dockerfile I use this:

```
COPY plugin/pom.xml .
RUN mvn dependency:go-offline
COPY plugin/src ./src

RUN mvn clean package
```

The `COPY` command calculates hash of the copied files and invalidates cache if they changed. So here's how it works:

1. We copy `pom.xml` that contains all the dependency information
2. We ask Maven to download all the jars for the build. If `pom.xml` changes, Docker cache will get invalidated before going into Maven Offline mode and will force it re-download everything. 
3. Then, we do similar trick with `plugin/src` and `mvn package`. 

I still think it's beneficial to use the ` --no-cache src` switch in the `uds-core-integration-tests` task to have an extra layer of protection against stale caches. But for iterating with the plugin, I managed to cut the rebuild time to just 2-3 seconds.

PR #441, opened 2025-04-21
fix: remove dynamic client registration

UnicornChance:
## Description
Keycloak dynamic client registration strategy is being deprecated in this release and this PR aims to remove the necessary pieces associated to that strategy and update documentation.

## Related Issue

Fixes #[1499](https://github.com/defenseunicorns/uds-core/issues/1499)

## Type of change

## Checklist before merging


tasks.yaml
```diff
@@ -148,7 +144,7 @@ tasks:
       - cmd: cd uds-core && uds run create:single-layer && uds run create:single-layer-callable --set LAYER=identity-authorization
       - task: grafana-group-auth-build-deploy
       - cmd: uds create bundles/k3d-core-slim-dev --confirm --no-progress
-      - cmd: uds deploy bundles/k3d-core-slim-dev/uds-bundle-k3d-core-slim-dev-*.tar.zst --set=core-identity-authorization.KEYCLOAK_CONFIG_IMAGE=uds-core-config:keycloak --set=core-base.PEPR_KEYCLOAK_CLIENT_STRATEGY=${PEPR_KEYCLOAK_CLIENT_STRATEGY} --confirm --no-progress
+      - cmd: uds deploy bundles/k3d-core-slim-dev/uds-bundle-k3d-core-slim-dev-*.tar.zst --set=core-identity-authorization.KEYCLOAK_CONFIG_IMAGE=uds-core-config:keycloak --confirm --no-progress
```
slaskawi, line 147: We probably also need to modify the list of trusted hosts in Keycloak to ensure all the requests are denied. Then, I think it would be great to leave just one Cypress tests that tries to invoke the Dynamic Client Registration (with a simple HTTP GET) and gets denied by Istio.

slaskawi (dismissed):

src/realm.json
```diff
@@ -1972,9 +1972,7 @@
                     "trusted-hosts": [
                         "127.0.0.6",
                         "*.${UDS_DOMAIN}",
-                        "*.${UDS_ADMIN_DOMAIN}",
```
slaskawi, line 1975: We probably also need some upgrade docs.

src/realm.json
```diff
@@ -1972,9 +1972,7 @@
...
-                        "*.pepr-uds-core-watcher.pepr-system.svc.cluster.local",
-                        "*.keycloak.svc.cluster.local"
+                        "*.${UDS_ADMIN_DOMAIN}"
```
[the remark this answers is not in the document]
UnicornChance: added this!

slaskawi (approved):

docs/reference/UDS Core/IdAM/upgrading-versions.md
```diff
@@ -4,11 +4,24 @@ title: Upgrading Versions
...
 <details open>
 <summary>Upgrade Details</summary>
 
+In uds-identity-config versions v0.14.0+, the UDS Identity Config has removed `Dynamic Client Registration`. Part of this means that we need to remove a couple of the `Trusted Hosts` configured in uds-identity-config. To manually configure this:
+   - Navigate to the `UDS` realm (be aware that in KC v26.2 the UI/UX of navigating realms has changed)
+   - Go to `Clients` > `Client Registration` > `Trusted Hosts`
+   - Remove the following Trusted Hosts:
+      - `*.keycloak.svc.cluster.local`
+      - `*.pepr-uds-core-watcher.pepr-system.svc.cluster.local`
+   - Click the `Save` button
```
mjnagel, line 18: Do we need to document the max-clients restriction here? I think that's important for restricting usage of this endpoint - although that may be dependent on my other comment...

src/realm.json
```diff
@@ -1972,9 +1984,7 @@
...
```
mjnagel, line 1987: I'm not actually sure that we need this list to have any trusted hosts anymore? 127.0.0.6 was the originating IP for mesh traffic in sidecar mode (operator in sidecar mode). The two domain entries were required based on our client policy around redirect uri matching a trusted host iirc. I think we could remove all of these safely (if so, we should update the docs as well).

UnicornChance: due to other short timeline work and capacity constraints, i would advocate for leaving this as is and adding the max-clients + max-clients upgrade section and add a backlog item so that we can get this over the finish line.

UnicornChance: Chatted in person and decided to remove the 127.0.0.6 entry from here, but kept the everything else so that users could utilize the admin gateway to do dynamic creation.

docs/reference/UDS Core/IdAM/upgrading-versions.md
UnicornChance: added, thanks!

docs/reference/UDS Core/IdAM/upgrading-versions.md
```diff
@@ -4,11 +4,30 @@ title: Upgrading Versions
...
 
 <details open>
 <summary>Upgrade Details</summary>
 
+In uds-identity-config versions v0.14.0+, the UDS Identity Config has removed `Dynamic Client Registration`. Part of this means that we need to remove a couple of the `Trusted Hosts` configured in uds-identity-config. To manually configure this:
+   - Navigate to the `UDS` realm (be aware that in KC v26.2 the UI/UX of navigating realms has changed)
+   - Go to `Clients` > `Client Registration` > `Trusted Hosts`
+   - Remove the following Trusted Hosts:
+      - `*.keycloak.svc.cluster.local`
+      - `*.pepr-uds-core-watcher.pepr-system.svc.cluster.local`
```
mjnagel, line 17: also remove 127.0.0.6
```suggestion
      - `*.pepr-uds-core-watcher.pepr-system.svc.cluster.local`
      - `127.0.0.6`
```

mjnagel (approved):

docs/reference/UDS Core/IdAM/upgrading-versions.md
UnicornChance: thanks, missed that

mjnagel (approved):

Issue #460, opened 2025-05-05
saml-group-membership-mapper is not an allowed protocol mapper


mjnagel:
https://github.com/defenseunicorns/uds-identity-config/commit/69b3388080d06ac7e61b361eea46f44244e4e435 introduced "hardening" around the default allowed client scopes and protocol mappers. This introduced an issue however for several apps as `saml-group-membership-mapper` was not included in the default list.

For adopters of certain UDS packages (notably [GitLab](https://github.com/defenseunicorns/uds-package-gitlab/blob/main/charts/config/templates/uds-package.yaml#L104-L113)) this caused a regression where SSO was not working as expected due to the client not being setup the same way.

This SAML group mapper should be included in the default list and we should review other UDS Packages to identify if there are any other missing defaults.

PR #461, opened 2025-05-05
fix: add grouplist to default allowed mappers

mjnagel:
## Description

Adds grouplist and `AudienceProtocolMapper` (tbd) to the default list of allowed mappers based on current package usage.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-identity-config/issues/460

## Type of change

## Checklist before merging

### DIFF
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
@@ -19,7 +19,6 @@ import org.keycloak.protocol.oidc.mappers.*;
 import org.keycloak.protocol.saml.mappers.RoleListMapper;
 import org.keycloak.protocol.saml.mappers.UserAttributeStatementMapper;
 import org.keycloak.protocol.saml.mappers.UserPropertyAttributeStatementMapper;
-import org.keycloak.protocol.saml.mappers.GroupMembershipMapper;
 import org.keycloak.representations.idm.ClientRepresentation;
 import org.keycloak.services.clientpolicy.ClientPolicyContext;
 import org.keycloak.services.clientpolicy.ClientPolicyException;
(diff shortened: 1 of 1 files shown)


src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -55,6 +61,8 @@ public class UDSClientPolicyPermissionsExecutor implements ClientPolicyExecutorP
             UserAttributeStatementMapper.PROVIDER_ID,
             UserAttributeMapper.PROVIDER_ID,
             UserPropertyAttributeStatementMapper.PROVIDER_ID,
+            GroupMembershipMapper.PROVIDER_ID,
+            AudienceProtocolMapper.PROVIDER_ID, // tbd: should we add this by default to support https://github.com/uds-packages/argo-workflows/blob/5257c03814a9f6b3925e26aaaea914918344dc37/chart/templates/uds-package.yaml#L26
```
mjnagel, line 65: I think I'm leaning towards adding this to support the current use-cases (this was found by searching for protocolMappers across our orgs). We could revisit this in the future potentially, but I'd rather make things work out of the box here for our users.
```suggestion
            AudienceProtocolMapper.PROVIDER_ID,
```

Curious on @slaskawi and @UnicornChance thoughts...

UnicornChance: I don't see why not. We're not supplying the configuration out of the box, so it will take a few extra hoops for someone to make use of this "allowed" aud client protocol. Two areas of concern are misconfiguration and technically its not least privileged configuration, but i really don't think either of those are good enough reasons to not add it.

slaskawi: Yep, let's add it!

slaskawi (approved):

noahpb (approved):

PR #471, opened 2025-05-09
chore: docker image fix

slaskawi:
## Description

Docker build fix (Chainguard changed image tags)

## Type of change

## Checklist before merging

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -8,7 +8,7 @@
 ###################################################################################
 FROM maven:3-eclipse-temurin-17-focal AS plugin
 
-WORKDIR /tmp
+WORKDIR /home/build
 
 COPY plugin/pom.xml .
 RUN mvn dependency:go-offline
@@ -39,7 +39,7 @@ FROM maven:3-eclipse-temurin-17-focal AS truststore
 USER root
 RUN apt-get update && apt-get install unzip -y
 
-WORKDIR /tmp
+WORKDIR /home/build
 
 ARG CA_REGEX_EXCLUSION_FILTER="\(.*EMAIL.*\|.*SW.*\)"
 ARG CA_ZIP_URL=https://dl.dod.cyber.mil/wp-content/uploads/pki-pke/zip/unclass-dod_approved_external_pkis_trust_chains.zip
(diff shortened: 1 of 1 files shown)


UnicornChance (approved):
good catch, i also ran into this. thanks for getting the PR up

UnicornChance (commented):
the image is currently wrong, its in the organizational images now. which i believe is fine. We'll just have to make note of that if/when we ever change to a different registry.

src/Dockerfile
```diff
@@ -6,7 +6,7 @@
 # Build the Keycloak java plugin and copy any extra-jars                          #
 #                                                                                 #
 ###################################################################################
-FROM cgr.dev/chainguard/maven:openjdk-17 AS plugin
+FROM cgr.dev/chainguard/maven:3-jdk17 AS plugin
```
UnicornChance, line 9: the only public images are the `latest` and `latest-dev`. Which is really unfortunate because we can't use private organization images in this repo. I tested locally and it does seem like using `latest` works, however that does bring some concerns around flakiness to tests and churn of the image updates.

noahpb: Another callout - we should only build with `jdk` and then set the runtime for the image to `jre`. Perhaps we could take an approach similar to [this](https://github.com/adorsys/keycloak-config-cli/blob/f9e365561231c40596428dcdad96024704fe492c/Dockerfile#L3).

slaskawi: I think I found a good candidate - https://hub.docker.com/layers/library/maven/3-eclipse-temurin-17-focal/images/sha256-4518b3124c6147a30de5d49f59dc3404e154b40c2d47841137d621596d6cc401

slaskawi: I went ahead and used the `maven:3-eclipse-temurin-21-jammy`, which is the latest Maven 3 with JDK 17 based on Ubuntu Jammy. 

The JDK 21 is requires as Keycloak is compiled against this precise version (see [here](https://github.com/keycloak/keycloak/blob/main/.github/actions/java-setup/action.yml#L12C6-L12C18)). We also need to baseline on Maven 3 as 4 hasn't been tested with this plugin so far.

Finally, we're specifically requiring Ubuntu Jammy, so that we have access to `apt-get`. This image has already OpenSSL and we just need to install `unzip` there. By reusing the image we shaved around 10-20 seconds of the build time. 

@noahpb To answer your JDK/JRE question, the UDS Identity Config is just a "holder" of files. The InitContainer copies them into proper placed before Keycloak bootstrap process starts. So it doesn't really matter what JDK/JRE is in this image, but it could what's in the Keycloak base image. That's the runtime that is being used.

UnicornChance (approved):

noahpb (approved):

PR #527, opened 2025-06-25
docs: update theme documentation, add doc shim

mjnagel:
## Description

Adds a CI docs shim workflow, allowing for docs changes to skip full testing and only run the shim.

Also significantly reworks the documentation around custom theming/branding/terms&conditions.

### DIFF
--- a/docs/reference/uds-core/idam/customization.md
+++ b/docs/reference/uds-core/idam/customization.md
@@ -61,7 +61,7 @@ packages:
                         name: keycloak-theme-overrides
 ```
 
-The configuration supports only four potential keys: `background.png`, `logo.png`, `footer.png`, and `favicon.png` which must exist in the corresponding ConfigMap(s). Note that you must pre-create this ConfigMap in the `keycloak` namespace before deploying/upgrading Core with these overrides. In the above example all images are in the same ConfigMap named `keycloak-theme-overrides`. An easy way to generate the ConfigMap manifest is using the following command (including whichever images you need and specifying the correct paths to your local images):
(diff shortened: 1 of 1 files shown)


docs/reference/uds-core/idam/customization.md
```diff
@@ -61,21 +61,42 @@ packages:
                         name: keycloak-theme-overrides
 ```
 
-The configuration supports only four potential keys: `background.png`, `logo.png`, `footer.png`, and `favicon.png` which are expected to exist in the corresponding ConfigMaps. In this example, all four images reside in the same ConfigMap named `keycloak-theme-overrides`. The values of these keys are base64 encoded images hosted as `binaryData` part of the ConfigMap. You can create it using the following command:
+The configuration supports only four potential keys: `background.png`, `logo.png`, `footer.png`, and `favicon.png` which must exist in the corresponding ConfigMap(s). Note that you must pre-create this ConfigMap in the `keycloak` namespace before deploying/upgrading Core with these overrides. In the above example all images are in the same ConfigMap named `keycloak-theme-overrides`. An easy way to generate the ConfigMap manifest is using the following command (including whichever images you need and specifying the correct paths to your local images):
```
slaskawi, line 64: > which must exist in the corresponding ConfigMap(s)

To me this reads as if user had to provide all 4 keys. Could we reword it to highlight that user may provide any of them? Might be background, might be favicon or might be all of them at once.

docs/reference/uds-core/idam/customization.md
```diff
@@ -61,21 +61,42 @@ packages:
...
   --from-file=footer.png=path/to/local/directory/footer.png \
-  --from-file=favicon.png=path/to/local/directory/favicon.png
+  --from-file=favicon.png=path/to/local/directory/favicon.png \
+  -n keycloak --dry-run=client -o=yaml > theme-image-cm.yaml
 ```
 
+To deploy this it is easiest to make a small zarf package referencing the manifest you just created:
+
+```yaml
+kind: ZarfPackageConfig
```
slaskawi, line 78: This is so clever! I love this approach!

slaskawi (approved):
LGTM with one nitpick comments. Please feel free to either address or ignore it.

docs/reference/uds-core/idam/customization.md
mjnagel: Let me know if this makes it more clear - https://github.com/defenseunicorns/uds-identity-config/pull/527/commits/e4200a20dd82a3890e419c2c225aadc0e1d21927

PR #541, opened 2025-07-08
feat: aws alb support

slaskawi:
## Description

This Pull Request introduces the support for a AWS ALB connected to the Tenant Gateway.

This Pull Request depends on https://github.com/defenseunicorns/uds-core/pull/1670

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1718
Fixes https://github.com/defenseunicorns/uds-core/issues/1721
Fixes https://github.com/defenseunicorns/uds-core/issues/1723
Relates to https://github.com/defenseunicorns/uds-core/issues/1630

## Type of change

## Checklist before merging

### DIFF
--- a/src/test/cypress/cypress.config.ts
+++ b/src/test/cypress/cypress.config.ts
@@ -5,7 +5,7 @@
 
 import { defineConfig } from "cypress";
 
-const useCAC = process.env.USE_CAC === "true";
+const useCAC = true;//process.env.USE_CAC === "true";
 
 module.exports = defineConfig({
   clientCertificates: useCAC
@@ -27,7 +27,7 @@ module.exports = defineConfig({
     setupNodeEvents(on, config) {
       // implement node event listeners here
     },
-    retries: 3,
+    retries: 0,
     specPattern: "e2e/**/*.cy.ts",
     supportFolder: "support/",
     supportFile: "support/e2e.ts",
--- a/src/test/cypress/cypress.config.ts
+++ b/src/test/cypress/cypress.config.ts
@@ -5,7 +5,7 @@
 
 import { defineConfig } from "cypress";
 
-const useCAC = true;//process.env.USE_CAC === "true";
+const useCAC = process.env.USE_CAC === "true";
 
 module.exports = defineConfig({
   clientCertificates: useCAC
--- a/src/test/cypress/cypress.config.ts
+++ b/src/test/cypress/cypress.config.ts
@@ -27,7 +27,7 @@ module.exports = defineConfig({
     setupNodeEvents(on, config) {
       // implement node event listeners here
     },
-    retries: 0,
+    retries: 3,
     specPattern: "e2e/**/*.cy.ts",
     supportFolder: "support/",
     supportFile: "support/e2e.ts",
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -2,8 +2,8 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
 includes:
-  - remote: https://raw.githubusercontent.com/defenseunicorns/uds-common/v1.16.3/tasks/lint.yaml
-  - common-setup: https://raw.githubusercontent.com/defenseunicorns/uds-common/v1.16.3/tasks/setup.yaml
+  - remote: https://raw.githubusercontent.com/defenseunicorns/uds-common/v1.16.4/tasks/lint.yaml
+  - common-setup: https://raw.githubusercontent.com/defenseunicorns/uds-common/v1.16.4/tasks/setup.yaml
   - setup: ./tasks/setup.yaml
 
 variables:
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -217,8 +217,6 @@ tasks:
         cmd: uds deploy bundles/external-loadbalancer/uds-bundle-external-loadbalancer-*.tar.zst --set=core-identity-authorization.KEYCLOAK_CONFIG_IMAGE=uds-core-config:keycloak --confirm --no-progress
       - description: "Recreate users"
         task: create-users
-      - description: "Restart the Tenant Gateway to pick the new Number of Trusted Proxies up"
-        cmd: kubectl rollout restart deployment tenant-ingressgateway -n istio-tenant-gateway
       - description: "Setup the infrastructure for the external load balancer"
         cmd: uds run setup:external-loadbalancer
       - description: "Run Cypress tests with the external load balancer"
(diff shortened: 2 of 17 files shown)


src/test/cypress/cypress.config.ts
```diff
@@ -5,7 +5,7 @@
 
 import { defineConfig } from "cypress";
 
-const useCAC = process.env.USE_CAC === "true";
+const useCAC = true;//process.env.USE_CAC === "true";
```
mjnagel, line 8: Was this change intentional?

tasks.yaml
```diff
@@ -125,7 +126,8 @@ tasks:
     description: "Integration Tests"
     actions:
       - task: cypress-tests
-      - task: theme-customization-tests
+#      - task: theme-customization-tests
```
mjnagel, line 129: Should this be re-added somewhere (or is it already) to make sure we don't lose coverage?

tasks/setup.yaml
```diff
@@ -0,0 +1,57 @@
...
+        cmd: |
+          yq -r '.tls.cert' uds-core/src/istio/values/config-tenant.yaml | base64 -d > tasks/haproxy.pem
+          yq -r '.tls.key' uds-core/src/istio/values/config-tenant.yaml | base64 -d > tasks/haproxy.pem.key
+          yq -r '.tls.cacert' src/test/cypress/certs/tls_cacert.yaml | base64 -d > tasks/haproxy-client-ca.pem
+      - description: "Sanity check HAProxy configuration"
+        cmd: |
+          TENANT_IP=$(uds zarf tools kubectl get svc tenant-ingressgateway -n istio-tenant-gateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}' || echo "unknown")
+          if [ "$TENANT_IP" != "172.18.0.201" ]; then
+            echo "\033[33m⚠️ Tenant Gateway external IP is not is not 172.18.0.201, got $TENANT_IP. The setup may not work correctly. \033[0m"
+          fi
```
mjnagel, line 32: When I deploy `uds-k3d` I think my IP range changes from time to time (always in the `172.x` range). Could we just dynamically set/change this in the HAProxy config? Potentially just set it to an env in the config file, then pass that env into `docker run`?

tasks/setup.yaml
```diff
@@ -0,0 +1,57 @@
...
+          docker run -d \
+            --name haproxy \
+            --network k3d-uds \
+            -v $(pwd)/tasks/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro \
+            -v $(pwd)/tasks/haproxy.pem:/etc/haproxy/certs/haproxy.pem:ro \
+            -v $(pwd)/tasks/haproxy.pem.key:/etc/haproxy/certs/haproxy.pem.key:ro \
+            -v $(pwd)/tasks/haproxy-client-ca.pem:/etc/haproxy/certs/haproxy-client-ca.pem:ro \
+            -p 80:80 \
+            -p 443:443 \
+            haproxy:3.2-alpine
```
mjnagel, line 47: We should probably ensure this is tracked with renovate.

mjnagel (commented):
Very clever setup here - just a few comments!

src/test/cypress/cypress.config.ts
slaskawi: No, sorry - it's a testing leftover.

tasks/setup.yaml
slaskawi: Done!

tasks/setup.yaml
slaskawi: This was the initial approach I started with. It tuned out this stands in the way when debugging and restarting HAProxy frequently. In this scenario you often have the k3d Load Balancer turned down and there's no way to obtain the $TENANT_IP. This was the primary reason I decided to "fix" it in the configuration.

But I've made a compromise - I'm checking the IP and if it's indeed different - I'm calling sed and updating it.

tasks/setup.yaml
mjnagel: May be missing a renovate match that is handling this, but the ones I am familiar with typically look for the version on the next line. Since the renovate comment here is separated from the image by several lines I'm not sure it will catch it?

You could probably use this match: https://github.com/defenseunicorns/uds-common/blob/main/config/renovate.json5#L148-L164

Something like:
```console
          # renovate: datasource=docker depName=haproxy versioning=docker
          HAPROXY_VERSION=3.2-alpine
          docker run -d \
            --name haproxy \
            --network k3d-uds \
            -v $(pwd)/tasks/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro \
            -v $(pwd)/tasks/haproxy.pem:/etc/haproxy/certs/haproxy.pem:ro \
            -v $(pwd)/tasks/haproxy.pem.key:/etc/haproxy/certs/haproxy.pem.key:ro \
            -v $(pwd)/tasks/haproxy-client-ca.pem:/etc/haproxy/certs/haproxy-client-ca.pem:ro \
            -p 80:80 \
            -p 443:443 \
            haproxy:${HAPROXY_VERSION}
```

tasks/setup.yaml
```diff
@@ -0,0 +1,58 @@
...
+          echo "Waiting for HAProxy to start..."
+          sleep 3
+      - description: "Sanity check for HAProxy"
+        cmd: |
+          STATUS_CODE=$(curl --max-time 10 -s --cert src/test/cypress/certs/test.pfx --cert-type P12 https://sso.uds.dev -o /dev/null -w "%{http_code}" -s || echo "unknown")
+          if [ "$STATUS_CODE" != "301" ]; then
+            echo "\033[33m⚠️ Expected HTTP 301, got $STATUS. The setup may not work correctly \033[0m"
+            echo "\033[33m⚠️ HAProxy logs \033[0m"
+            docker logs haproxy || true
+          fi
```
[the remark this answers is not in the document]
slaskawi: Good idea! Done!

tasks/setup.yaml
slaskawi: Done!

tasks.yaml
```diff
@@ -144,7 +146,7 @@ tasks:
     description: "Clone UDS Core for integration testing"
     actions:
       - cmd: rm -rf uds-core
-      - cmd: git clone --branch ${CORE_VERSION} https://github.com/defenseunicorns/uds-core.git
+      - cmd: git clone --branch 1630-ALB-Support https://github.com/defenseunicorns/uds-core.git
```
mjnagel, line 149: Will want to revert when the core branch is merged.

mjnagel (commented):
LGTM - holding off on approval since this requires the core side to be merged to main first.

tasks.yaml
slaskawi: Switched to main for the time-being (will revert back after the release)

chance-coleman (approved):

Issue #532, opened 2025-06-27
Skip robot detection when using mobile


slaskawi:
## Summary

With the new Keycloak theme, the registration screen with mobile devices is very clunky. It requires touching screen and swiping. This is not the best UX.

We might want to consider either removing the robot detection mechanism or speeding it up for mobiles.

PR #556, opened 2025-07-22
feat: faster robot detection for mobiles

slaskawi:
## Description

This Pull Request speeds up the robot detection at the registration page which is painfully slow on mobile. 

## Related Issue

Fixes https://github.com/defenseunicorns/uds-identity-config/issues/532

## Type of change

## Checklist before merging


src/theme/login/register.ftl
```diff
@@ -309,8 +309,11 @@
         document.getElementById('rank').value = "${(register.formData['rank']!'')}";
     }
     (function() {
-        const threshold = 250;
         let count = 0;
+        // Mobile device detection
+        const isMobile = /Mobi|Android|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent);
```
mjnagel, line 314: I think this is probably fine but wondering if we should align with how we detect mobile in the CSS side (via [browser width](https://github.com/defenseunicorns/uds-identity-config/blob/532-speed_robot_detection_on_mobiles_up/src/theme/login/resources/css/uds.css#L805-L806))? I'm not sure if some of the "slow robot checking" is related to the CSS/theme changes that happen on smaller screens, but seems sane to keep these aligned unless there's an issue with checking width on the JS side here.

```suggestion
        // Mobile device detection
        const isMobile = window.innerWidth < 768;
```

I think it would just be like this?

PR #652, opened 2025-10-03
feat: always log groups in Admin Events related to users

slaskawi:
## Description

...

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/eventListeners/JSONLogEventListenerProvider.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/eventListeners/JSONLogEventListenerProvider.java
@@ -5,22 +5,45 @@
 
 package com.defenseunicorns.uds.keycloak.plugin.eventListeners;
 
+import java.lang.invoke.MethodHandles;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.util.Iterator;
+import java.util.Map;
 import java.util.Map.Entry;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/eventListeners/JSONLogEventListenerProvider.java
```diff
@@ -40,9 +63,33 @@ public void close() {
...
+                    String userId = userIdAsObject.toString();
+                    RealmModel realm = session.realms().getRealm(adminEvent.getRealmId());
+                    UserModel user = session.users().getUserById(realm, userId);
+                    List<String> groups = user.getGroupsStream()
+                            .map(ModelToRepresentation::buildGroupPath)
+                            .collect(Collectors.toCollection(ArrayList::new));
+                    logger.debug("groups: {}", groups);
+                    representationAsMap.putIfAbsent("groups", groups);
+                }
+                adminEvent.setRepresentation(objectMapper.writeValueAsString(representationAsMap));
```
chance-coleman, line 86: Right now you always call setRepresentation, even if `userIdAsObject` was missing.That means you might re-serialize and replace the JSON unnecessarily. Could that create subtle issues if other fields are reordered or formatting differs?

chance-coleman (dismissed):
this looks good to me, one question about implementation and for my own understanding.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/eventListeners/JSONLogEventListenerProvider.java
slaskawi: That's a good point! Let me optimize this bit!

Issue #710, opened 2025-11-07
Expose a switch to disable the "username" field in the self-registration screen


slaskawi:
## Summary

Similarly to https://github.com/defenseunicorns/uds-identity-config/issues/682, we'd like to disable the "username" field in the self-registration screen. 

The field should be hidden by default and there should be exposed an option to turn it on. The primary motivation for this is that some Missions may not want to join user profiles during the First Broker Login flow. In this (quite niche but still) case, we need to have usernames.

PR #714, opened 2025-11-13
feat: configurable username in user self-registration

slaskawi:
## Description

This Pull Request introduces a configuration option to login users with emails (in other words to treat emails as logins). 

This setting is configured at the Realm level and propagates to the User Self-Registration screen where login is replaced with email. 

⚠️  This change requires https://github.com/defenseunicorns/uds-core/pull/2134

## Related Issue

Fixes https://github.com/defenseunicorns/uds-identity-config/issues/710

## Type of change

## Checklist before merging

### DIFF
--- a/bundles/k3d-core-slim-dev/uds-bundle.template.yaml
+++ b/bundles/k3d-core-slim-dev/uds-bundle.template.yaml
@@ -14,7 +14,7 @@ packages:
   - name: init
     repository: ghcr.io/zarf-dev/packages/init
     # renovate: datasource=github-tags depName=zarf-dev/zarf versioning=semver
-    ref: v0.65.1
+    ref: v0.66.0
   - name: keycloak-identity-config
     path: ../../
     ref: dev


src/test/cypress/support/commands.ts
```diff
@@ -78,6 +78,18 @@ Cypress.Commands.add("loginUser", (username: string, password: string) => {
   cy.get("#kc-login").should("be.visible").click();
 });
 
+/**
+ * Supply the login page user creds (specifically email) and attempt to login
+ */
+Cypress.Commands.add("loginUserWithEmail", (email: string, password: string) => {
```
chance-coleman: I understand the addition here but i don't think its necessary when we have the `loginUser` command above and the only difference is the name of the parameter. The way i look at this PR is that we are using the email of a user as their username, so it's still a username just locked down to only being a email syntax.

chance-coleman (commented):
tested this locally, everything is looking good. couple comments on the cypress testing.

chance-coleman (dismissed):
LGTM, once uds-core PR is merged and task is pointed at main

PR #715, opened 2025-11-13
feat: group protection for service accounts

slaskawi:
## Description

This pull request includes an ADR that outlines three potential options for providing Group Protection for Service Accounts in UDS, along with the selected solution and an explanation of its consequences.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/2132

## Type of change

## Checklist before merging

### DIFF
--- a/docs/dev/adr/0004-group-protection-for-the-service-accounts.md
+++ b/docs/dev/adr/0004-group-protection-for-the-service-accounts.md
@@ -46,6 +46,7 @@ Advantages:
 Disadvantages:
 
 - Extends the existing Group/Permission model of the UDS Core Platform, which requires re-evaluating the Threat Model
+- There's no guarantee that proposed groups would cover the user application groups hierarchy
 
 3. Extend the existing Package CR `sso` spec with a protection mechanism based on custom claims
 


docs/dev/adr/0004-group-protection-for-the-service-accounts.md
```diff
@@ -0,0 +1,68 @@
...
+
+- Fine-level control over Service Account permissions
+
+Disadvantages:
+
+- Extends the existing Group/Permission model of the UDS Core Platform, which requires re-evaluating the Threat Model
+
+3. Extend the existing Package CR `sso` spec with a protection mechanism based on custom claims
+
+The implementation assumes extending the Package CR `sso` spec with a new mechanism custom claims checks, such as `azp` (that contains the name of the Service Account) or `preferred_username` (usually `service-account-<client-id>`). The UDS Core would adjust the `RequestAuthentication` to inspect the `@request.auth.claims.groups` (as described in the [Istio documentation](https://istio.io/latest/docs/tasks/security/authentication/jwt-route/#configuring-ingress-routing-based-on-jwt-claims)) and grant request access based on the presence of specific custom claims.
```
mjnagel: Is it accurate that this would only handle our custom plugin group protection, but doesn't help for end user apps that might specifically require the group claim?

docs/dev/adr/0004-group-protection-for-the-service-accounts.md
```diff
@@ -0,0 +1,68 @@
...
+
+- No additional changes required
+
+Disadvantages:
+
+- Requires two-step configuration (adding Protocol Mapper to allow list and then configuring it on the Package CR)
+
+2. Introduce the `/UDS Core/Admin/Service Account` and `/UDS Core/Auditor/Service Account` groups, and automatically add the appropriate `groups` claim to Service Account tokens
+
+This approach relies on extending the existing Group/Permission model in UDS Core down to Service Accounts. With this approach every Service Account that belongs `/UDS Core/Auditor/Service Account` is automatically member of the `/UDS Core/Auditor` group which makes the Package CR configuration simpler. An additional Protocol Mapper would be automatically added to every Keycloak Client that has Service Account enabled, which would add the `groups` claim with proper values based on the Service Account group membership.
```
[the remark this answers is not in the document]
slaskawi: Yes and no.

The idea here would be to create our own protocol mapper that would put these claims into the token. Please note, that you'd have limited options and you couldn't exploit it to produce an ordinary claim. It would be just used for the `group` and would produce one of two options.

docs/dev/adr/0004-group-protection-for-the-service-accounts.md
slaskawi: That is true - I added a Disadvantage about this. Please check if it looks good and resolve this thread.

slaskawi:
> For option 3, what happens for apps that require a groups claim for SA tokens? I think this would be out of scope for this option if im understanding it correctly.

@chance-coleman @mjnagel asked very similar question [here](https://github.com/defenseunicorns/uds-identity-config/pull/715#discussion_r2523968301). Unfortunately the option no 3 doesn't provide a solution to that problem. I've altered the text and added this as a disadvantage. 

> How frequently do we expect this to be used, wondering if engineering something like option 3 is worth the effort of changing the Package schema, managing security and misconfigurations or overly broad configurations if we allow for regex.

Perhaps @bburky or @ntwkninja know the answer. I've heard about one or two Mission apps that need it, so it's not a ton but frequently enough so that we should provide some option that enables this functionality. 

> I think my slight leaning is option 1, with an explicit "here be dragons" warning that users would be taking responsibility for their use of protocol mappers.

That was my second-best choice (I'm personally a fan of option no 3 - but as we clarified, the option 3 doesn't help applications that just need the `groups` claim).

@mjnagel @chance-coleman @joelmccoy I'll use option no 1 for the "decision" paragraph and send this document to a broader review.

slaskawi:
@bburky I'm fully onboard with all what you said. You're also correct that the lack of support adding the `groups` claim to the apps that need it is the primary reason we rejected option no 3 (which is still my personal favorite but sadly, doesn't fulfill all the cases).

slaskawi:
> Assuming the group names can be customized though... maybe this works? Most off the shelf apps do at least seem to allow specifying multiple groups in their config, so you could just additionally list this group too. Maybe some apps don't support multiple groups? This maybe probably works?

@bburky In my opinion, we'd end up very close to the option no 1 than.

Issue #722, opened 2025-11-17
More descriptive account lockout options


chance-coleman:
### Is your feature request related to a problem? Please describe.

There is two use cases when a user fails to login to their account, they can be permanently locked out or temporarily locked out. Both of these scenarios currently result in the error being shown to the user `Invalid username or password` which doesn't provide a lot of information.

In some environments this is considered a security concern to show this level of information.

### Describe the solution you'd like

When a user fails to login to a permanently locked account the message should inform them they need to reach out to an administrator for help (only way to unlock the account).

When a user fails to login to a temporarily locked account the message should inform them they are currently locked out from failed attempts and should try again later or reach out to an administrator.

This should be configurable in uds-core from a bundle and should default to false, meaning it's not enabled by default.

PR #723, opened 2025-11-17
feat: expose temporary lockouts and document

### LINKED ISSUE #691: Evaluate Permanent Lockout of Users
## Current Behavior
Currently when users fail to authenticate 3 times they are permanetly locked out of their account until an admin intervenes. Is this a requirement? Are there any STIGS that support this requirement or can we implement timeout and unlock users accounts a couple times before permanent lockout?

## Related Configurations
[Max 3 Auth Attempts](https://github.com/defenseunicorns/uds-identity-config/blob/main/src/realm.json5#L52)
[Permanent Lockout](https://github.com/defenseunicorns/uds-identity-config/blob/main/src/realm.json5#L45)
[Max number of lockouts before permanent lockout](https://github.com/defenseunicorns/uds-identity-config/blob/main/src/realm.json5#L46)

## How it works currently
- `failureFactor`: Number of consecutive failures before a lockout occurs.
- `permanentLockout` + `maxTemporaryLockouts`:
  - If `permanentLockout` is true, after `maxTemporaryLockouts` temporary lockouts, the user is permanently locked.
  - In  config, `maxTemporaryLockouts`: 0 means it will go straight to permanent lock once the threshold is met.
- Timing controls (if not permanent):
  - `waitIncrementSeconds`, `minimumQuickLoginWaitSeconds`, and `maxFailureWaitSeconds` control the backoff/lockout duration.

chance-coleman:
## Description

Expose the ability for configuring keycloak to allow temporary lockout before permanent lockout. Core continues to default to permanent lockout. However configuring a bundle with:
```
overrides:
  keycloak:
    keycloak:
      values:
        - path: realmInitEnv
          value:
            MAX_TEMPORARY_LOCKOUTS: 3
```
will allow for 3 temporary 15 minute lockouts of a users account before their account is permanent locked and an administrator needs to step in. This number is completely configurable.

Additionally added documentation for:
- different settings in the realm.json
- how to configure from bundle
- UDS Core defaults
- Manual steps to configure this in an existing realm
- Compliance implications if changing certain fields

### Out of Scope
Figuring out how to display more context on the locked out page. This is seen in some environments as being too much information because it confirms that an account exists, narrowing the scope for malicious actors. This would be configurable and default to off. [Issue](https://github.com/defenseunicorns/uds-identity-config/issues/722) to track this future work.

## Related Issue

Fixes #691 

Related Core PR

## Type of change

## Checklist before merging

### DIFF
--- a/docs/reference/uds-core/idam/account-lockout.md
+++ b/docs/reference/uds-core/idam/account-lockout.md
@@ -28,14 +28,14 @@ overrides:
 
 #### 1. Default UDS Core behavior: **permanent lockout** (`MAX_TEMPORARY_LOCKOUTS = 0`)
 
-1. User fails login **3 times** within a 15 minute window.
+1. User fails login **3 times** within a 12 hour window.
 1. Keycloak applies a **permanent lockout** once the threshold is hit.
 1. The account remains locked until an administrator manually unlocks it.
 
 #### 1. Optional: temporary then permanent mode (`MAX_TEMPORARY_LOCKOUTS > 0`)
 
 1. Configure `MAX_TEMPORARY_LOCKOUTS` to a non‑zero value.
-1. User fails login **3 times within the 15 minute window** → Keycloak applies a **temporary lockout** for 15 minutes.
+1. User fails login **3 times within the 12 hour window** → Keycloak applies a **temporary lockout** for 15 minutes.
 1. After the temporary lockout expires, the account unlocks.
 1. If the user triggers more temporary lockouts than allowed by the configured `MAX_TEMPORARY_LOCKOUTS` value, Keycloak escalates the account to a **permanent lockout**. The user has **up to 12 hours** during which additional lockouts count toward this limit. For example:
 
--- a/src/realm.json5
+++ b/src/realm.json5
@@ -42,15 +42,15 @@
     "resetPasswordAllowed": "${USERNAME_PASSWORD_AUTH_ENABLED:true}",
     "editUsernameAllowed": false,
     "bruteForceProtected": true,
-    "permanentLockout": true,                                     // allow Keycloak to permanently lock the account once temporary lockout limit is exceeded
-    "maxTemporaryLockouts": "${REALM_MAX_TEMPORARY_LOCKOUTS:0}",  // number of temporary lockouts allowed before permanent lockout
-    "bruteForceStrategy": "MULTIPLE",                             // uses incremental wait times based on multiple failed attempts
-    "maxFailureWaitSeconds": 86400,                               // maximum wait time Keycloak can apply (24 hours)
-    "minimumQuickLoginWaitSeconds": 60,                           // minimum delay when rapid login failures are detected
-    "waitIncrementSeconds": 900,                                  // base temporary lockout duration after reaching failureFactor (15 minutes)
-    "quickLoginCheckMilliSeconds": 1000,                          // time window for detecting rapid failed attempts (1 second)
-    "maxDeltaTimeSeconds": 43200,                                 // failure counting window (12 hours)
-    "failureFactor": 3,                                           // number of failed attempts required to trigger a lockout
+    "permanentLockout": true,                                     // When true, permanently disables the user once the permanent threshold is reached; admin must re-enable.
+    "maxTemporaryLockouts": "${REALM_MAX_TEMPORARY_LOCKOUTS:0}",  // Number of temporary lockouts allowed before permanent lockout; 0 means no temporary lockouts—reaching failureFactor can immediately trigger permanent lockout when permanentLockout is true.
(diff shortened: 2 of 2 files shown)


src/realm.json5
```diff
@@ -42,14 +42,15 @@
...
-    "maxDeltaTimeSeconds": 43200,
-    "failureFactor": 3,
+    "permanentLockout": true,                                     // allow Keycloak to permanently lock the account once temporary lockout limit is exceeded
+    "maxTemporaryLockouts": "${REALM_MAX_TEMPORARY_LOCKOUTS:0}",  // number of temporary lockouts allowed before permanent lockout
+    "bruteForceStrategy": "MULTIPLE",                             // uses incremental wait times based on multiple failed attempts
+    "maxFailureWaitSeconds": 86400,                               // maximum wait time Keycloak can apply (24 hours)
+    "minimumQuickLoginWaitSeconds": 60,                           // minimum delay when rapid login failures are detected
+    "waitIncrementSeconds": 900,                                  // base temporary lockout duration after reaching failureFactor (15 minutes)
+    "quickLoginCheckMilliSeconds": 1000,                          // time window for detecting rapid failed attempts (1 second)
+    "maxDeltaTimeSeconds": 43200,                                 // failure counting window (12 hours)
```
mjnagel: Is there a reason this is so high (as compared to compliance requirements of 15 mins)? Looks like we're maintaining what was there, and this would technically be _stricter_ but I am curious if we should lower the delta time for usability?

chance-coleman: yeah good call, i reverted that back to 15 minutes.

chance-coleman: `maxDeltaTimeSeconds` must be longer than `waitIncrementSeconds`, or permanent lockout will never happen.

If both values are the same (e.g., 15 minutes), the temporary lock expires at the same moment the failure history resets.

That means Keycloak treats every new set of failures like the first time, so permanent lockout never triggers.

We need to make `maxDeltaTimeSeconds` much bigger than `waitIncrementSeconds` so multiple lockouts can accumulate and trigger a permanent lock.

docs/reference/uds-core/idam/account-lockout.md
```diff
@@ -0,0 +1,109 @@
...
+
+* NIST SP 800‑53 AC‑7 requires defining and enforcing a lockout policy. Both supported UDS Core options meet this requirement.
+
+### Compliance impact if changed if changed
+
+* `failureFactor` > 3 violates STIG expectations.
+* `maxDeltaTimeSeconds` < 900 breaks the “3 attempts in 15 minutes” requirement.
+* `waitIncrementSeconds` < 900 non‑compliant with the 15‑minute lockout minimum.
+* Increasing `maxTemporaryLockouts` changes the escalation behavior; allowed but must be documented.
+* Turning off permanent lockout is allowed, but alters the operational model and must be documented.
```
mjnagel: We may just want to leave the details here to Lula mappings and compliance justifications, and note something more generic like "There may be compliance impacts when modifying lockout behavior, be aware of any environment specific requirements (NIST controls or STIG requirements) for brute-force protection."

src/realm.json5
chance-coleman: @mjnagel  @slaskawi To recap async conversations + some [upstream keycloak code](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/services/managers/DefaultBruteForceProtector.java). The `maxDeltaTimeSeconds` is resused for two different configurations that in our case affects how we want to configure things.
  1. control the amount of time a user has to successfully login
    - E.g. a user has 3 failed login attempts in 15 mintues
  2. control the amount of time keycloak will keep track of temporary lockouts, resulting in a permanent lockout
    - E.g. 3 temporary lockouts in 6 hours results in a permanent lockout
    
This overlapping use of this setting restricts us from being able to configure the following use case:
  1. user fails 3 login attempts in 15 minutes resulting in a 15 minute temporary lockout
  2. upon lockout completion user fails another 3 login attempts in 15 minutes resulting in a permanent lockout

**How does it restrict us?** Because the `maxDeltaTimeSeconds` can't set both of the earlier configurations independently where we expect 3 failed attempts to happen in 15 minutes window **and** the number of temporary lockouts that result in a permanent lockout to happen in a 12 hour window.

Looking at STIGS and NIST/DOD compliance docs it seems that we have two requirements for hardening core by default: 3 invalid attempts in a 15 minute window. Anything more than 15 minute window would be only making that requirement more secure (albiet less UX friendly). 

So i'm proposing that we keep a long period of time for the attempts window so that it also encompasses a number of temporary lockouts, so that we can safely result in a permanent lockout. Open to suggestions for how that window should be, i like 12 hours (currently set to this) but could see 6 or 24 as well. One thing to keep in mind is that if we set a smaller window and an admin configures keycloak to have a high number of temporary lockouts before permanent lockout (i.e. 50) they may never reach the point of permanent lockout (maybe this is a feature?).

PR #838, opened 2026-04-15
feat(CORE-152): re-introduce the Signed JWTs to the UDS Operator

slaskawi:
## Description

This Pull Request attempts to re-introduce the Signed JWTs into the UDS Core with a custom - UDS Identity Provider.

The approach in this PR differs from the upstream as the upstream assumes Kubernetes **IS** the IdP that issues Service Account Tokens. This doesn't have to be the truth in case of managed Kubernetes installation that may use a centralized IdP (e.g. Azure will use Entra etc). 

If such a 3rd party IdP integration is used, the Kubernetes JWKS url correctly points the keys (so the signature validation is not impacted) but the OIDC Discovery endpoint points to that 3rd party:

```
$ kubectl get --raw /.well-known/openid-configuration | jq
{
  "issuer": "https://oidc.eks.<MASKED>.amazonaws.com/id/<MASKED>",
  "jwks_uri": "https://oidc.eks.<MASKED>amazonaws.com/id/<MASKED>/keys",
  "response_types_supported": [
    "id_token"
  ],
  "subject_types_supported": [
    "public"
  ],
  "id_token_signing_alg_values_supported": [
    "RS256"
  ]
}
```

This means, we need to take a different approach - the Client Assertions validation need to first load the OIDC discovery endpoint from Kubernetes, get the issuer and then proceed with the validation based on that obtained issuer. 

## Related Issue

Relates to Relates to https://linear.app/defense-unicorns/issue/CORE-152/use-signed-jwts-for-uds-operator

## Type of change

## Checklist before merging

### DIFF
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
@@ -18,6 +18,7 @@ import org.keycloak.models.IdentityProviderModel;
 import com.defenseunicorns.uds.keycloak.plugin.broker.kubernetes.UDSKubernetesIdentityProviderFactory;
 
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
 import org.jboss.logging.Logger;
@@ -91,8 +92,9 @@ public class UDSClientAssertionStrategy implements ClientAssertionStrategy {
         }
         var idpStorage = context.getSession().identityProviders();
 
-        List<LookupResult> matches = context.getRealm().getClientsStream()
-            .filter(c -> subject.equals(c.getAttribute(FederatedJWTClientAuthenticator.JWT_CREDENTIAL_SUBJECT_KEY)))
+        List<LookupResult> matches = context.getSession().clients()
+            .searchClientsByAttributes(context.getRealm(),
+                Map.of(FederatedJWTClientAuthenticator.JWT_CREDENTIAL_SUBJECT_KEY, subject), null, null)
             .map(matchingClient -> {
                 String idpAlias = matchingClient.getAttribute(FederatedJWTClientAuthenticator.JWT_CREDENTIAL_ISSUER_KEY);
                 if (idpAlias == null) {
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Defense Unicorns
+ * Copyright 2026 Defense Unicorns
  * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
  */
 
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
@@ -106,7 +106,7 @@ public class UDSClientAssertionStrategy implements ClientAssertionStrategy {
                 }
                 return null;
             })
-            .filter(r -> r != null)
+            .filter(Objects::nonNull)
             .findFirst()
             .orElse(null);
     }
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2025 Defense Unicorns
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+ */
+
+package com.defenseunicorns.uds.keycloak.plugin.authentication.authenticators.client;
+
+import org.keycloak.OAuth2Constants;
+import org.keycloak.authentication.ClientAuthenticationFlowContext;
+import org.keycloak.authentication.authenticators.client.ClientAssertionState;
+import org.keycloak.authentication.authenticators.client.DefaultClientAssertionStrategy;
+import org.keycloak.authentication.authenticators.client.FederatedJWTClientAuthenticator;
+import org.keycloak.broker.provider.ClientAssertionIdentityProviderFactory.ClientAssertionStrategy;
+import org.keycloak.broker.provider.ClientAssertionIdentityProviderFactory.LookupResult;
+import org.keycloak.models.ClientModel;
+import org.keycloak.models.IdentityProviderModel;
+
+import com.defenseunicorns.uds.keycloak.plugin.broker.kubernetes.UDSKubernetesIdentityProviderFactory;
+
+import org.jboss.logging.Logger;
+
+/**
+ * Issuer-agnostic client assertion strategy for managed Kubernetes clusters.
+ *
+ * <h2>Why this exists</h2>
+ *
+ * The upstream {@link DefaultClientAssertionStrategy} looks up identity providers by matching
+ * the JWT assertion's {@code iss} claim against configured IdP issuer URLs. On self-hosted
+ * clusters (e.g. k3d), the SA token issuer is {@code https://kubernetes.default.svc.cluster.local}
+ * which matches the configured IdP — so the default strategy works.
+ *
+ * On managed Kubernetes clusters (EKS, AKS, GKE), the SA token issuer is a cloud-specific URL
+ * (e.g. {@code https://oidc.eks.us-gov-west-1.amazonaws.com/id/...}) that does NOT match the
+ * configured IdP issuer. The default strategy returns null, and authentication fails with
+ * {@code client_not_found} before our custom IdP's {@code verifyClientAssertion} is ever called.
+ *
+ * <h2>How the fallback works</h2>
+ *
+ * When the default issuer-based lookup fails, this strategy uses a client-first approach:
+ * <ol>
+ *   <li>Scan realm clients for one whose {@code jwt.credential.sub} matches the JWT subject</li>
+ *   <li>Read that client's {@code jwt.credential.issuer} attribute (which is the IdP alias)</li>
+ *   <li>Look up the IdP by alias via {@code IdentityProviderStorageProvider.getByAlias()}</li>
+ *   <li>Verify the IdP uses our {@code uds-kubernetes} provider type and is enabled</li>
+ * </ol>
+ *
+ * <h2>Why we don't use {@code getIdentityProvidersStream()}</h2>
+ *
+ * In Keycloak 26.5, {@code RealmModel.getIdentityProvidersStream()} delegates to
+ * {@code IdentityProviderStorageProvider.getAllStream(IdentityProviderQuery.userAuthentication())},
+ * which only returns {@code USER_AUTHENTICATION} type IdPs. Our {@code uds-kubernetes} IdP has
+ * type {@code CLIENT_ASSERTION}, so it is excluded from that stream. Using
+ * {@code getByAlias()} bypasses this filter entirely.
+ *
+ * @see DefaultClientAssertionStrategy
+ * @see UDSKubernetesIdentityProvider
+ */
+public class UDSClientAssertionStrategy implements ClientAssertionStrategy {
+
+    private static final Logger LOGGER = Logger.getLogger(UDSClientAssertionStrategy.class);
+
+    private final DefaultClientAssertionStrategy defaultStrategy = new DefaultClientAssertionStrategy();
+
+    @Override
+    public boolean isSupportedAssertionType(String assertionType) {
+        return OAuth2Constants.CLIENT_ASSERTION_TYPE_JWT.equals(assertionType);
+    }
+
+    @Override
+    public LookupResult lookup(ClientAuthenticationFlowContext context) throws Exception {
+        // 1. Try standard issuer-based lookup (handles k3d and matching-issuer cases)
+        LookupResult defaultResult = defaultStrategy.lookup(context);
+        if (defaultResult != null
+                && defaultResult.identityProviderModel() != null
+                && defaultResult.identityProviderModel().isEnabled()
+                && defaultResult.clientModel() != null) {
+            LOGGER.debug("Standard issuer-based lookup succeeded");
+            return defaultResult;
+        }
+
+        // 2. Fallback: client-first lookup for managed K8s clusters where issuer doesn't match
+        LOGGER.debug("Standard issuer-based lookup failed, falling back to client-first lookup");
+        ClientAssertionState state = context.getState(ClientAssertionState.class, ClientAssertionState.supplier());
+        String subject = state != null && state.getToken() != null ? state.getToken().getSubject() : null;
+        if (subject == null) {
+            LOGGER.debug("Fallback lookup skipped because client assertion subject is missing");
+            return null;
+        }
+        var idpStorage = context.getSession().identityProviders();
+
+        return context.getRealm().getClientsStream()
+            .filter(c -> subject.equals(c.getAttribute(FederatedJWTClientAuthenticator.JWT_CREDENTIAL_SUBJECT_KEY)))
+            .map(matchingClient -> {
+                String idpAlias = matchingClient.getAttribute(FederatedJWTClientAuthenticator.JWT_CREDENTIAL_ISSUER_KEY);
+                if (idpAlias == null) {
+                    return null;
+                }
+
+                IdentityProviderModel idp = idpStorage.getByAlias(idpAlias);
+                if (idp != null
+                        && UDSKubernetesIdentityProviderFactory.PROVIDER_ID.equals(idp.getProviderId())
+                        && idp.isEnabled()) {
+                    LOGGER.debugf("Fallback lookup matched client '%s' to IdP '%s'",
+                        matchingClient.getClientId(), idpAlias);
+                    return new LookupResult(matchingClient, idp);
+                }
+                return null;
+            })
+            .filter(r -> r != null)
+            .findFirst()
+            .orElse(null);
+    }
+}
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
@@ -17,6 +17,8 @@ import org.keycloak.models.IdentityProviderModel;
 
 import com.defenseunicorns.uds.keycloak.plugin.broker.kubernetes.UDSKubernetesIdentityProviderFactory;
 
+import java.util.Objects;
+
 import org.jboss.logging.Logger;
 
 /**
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
@@ -17,6 +17,7 @@ import org.keycloak.models.IdentityProviderModel;
 
 import com.defenseunicorns.uds.keycloak.plugin.broker.kubernetes.UDSKubernetesIdentityProviderFactory;
 
+import java.util.List;
 import java.util.Objects;
 
 import org.jboss.logging.Logger;
@@ -90,7 +91,7 @@ public class UDSClientAssertionStrategy implements ClientAssertionStrategy {
         }
         var idpStorage = context.getSession().identityProviders();
 
-        return context.getRealm().getClientsStream()
+        List<LookupResult> matches = context.getRealm().getClientsStream()
             .filter(c -> subject.equals(c.getAttribute(FederatedJWTClientAuthenticator.JWT_CREDENTIAL_SUBJECT_KEY)))
             .map(matchingClient -> {
                 String idpAlias = matchingClient.getAttribute(FederatedJWTClientAuthenticator.JWT_CREDENTIAL_ISSUER_KEY);
@@ -109,7 +110,14 @@ public class UDSClientAssertionStrategy implements ClientAssertionStrategy {
                 return null;
             })
             .filter(Objects::nonNull)
-            .findFirst()
-            .orElse(null);
+            .toList();
+
+        if (matches.size() > 1) {
+            throw new IllegalStateException(String.format(
+                "Ambiguous client assertion: subject '%s' matched %d clients. "
+                + "Each service account subject must map to exactly one client.", subject, matches.size()));
+        }
+
+        return matches.isEmpty() ? null : matches.get(0);
     }
 }
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
@@ -1,0 +1,158 @@
+/*
+ * Copyright 2025 Defense Unicorns
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+ */
+
+package com.defenseunicorns.uds.keycloak.plugin.authentication.authenticators.client;
+
+import org.keycloak.OAuth2Constants;
+import org.keycloak.authentication.AuthenticationFlowError;
+import org.keycloak.authentication.ClientAuthenticationFlowContext;
+import org.keycloak.authentication.authenticators.client.FederatedJWTClientValidator;
+import org.keycloak.models.ClientModel;
+import org.keycloak.representations.JsonWebToken;
+
+import org.jboss.logging.Logger;
+
+/**
+ * Custom JWT client validator that relaxes the {@code client_id} form parameter check.
+ *
+ * <h2>Why this exists</h2>
+ *
+ * The upstream {@link FederatedJWTClientValidator} inherits from {@code AbstractJWTClientValidator}
+ * which has a {@code private} method {@code validateClient()} that rejects requests where the
+ * {@code client_id} form parameter doesn't match the JWT {@code sub} claim. In the Kubernetes
+ * service account flow, the JWT subject is the SA identity
+ * (e.g. {@code system:serviceaccount:pepr-system:pepr-uds-core}) while the {@code client_id}
+ * parameter is the Keycloak client name (e.g. {@code uds-operator}). These intentionally differ.
+ *
+ * <p>Because {@code validateClient()} is private, we cannot override just that method. We must
+ * override {@code validate()} entirely and copy the private helper methods, modifying only the
+ * client_id check to log a debug message instead of failing.
+ *
+ * <p>This is a stop-gap solution until
+ * <a href="https://github.com/keycloak/keycloak/pull/48026">https://github.com/keycloak/keycloak/pull/48026</a>
+ * is merged upstream, which will make validation methods extensible.
+ *
+ * <h2>What changed vs upstream</h2>
+ *
+ * Only one behavioral change in {@code validateClientInternal()}:
+ * <pre>
+ * // Upstream (fails):
+ * if (clientIdParam != null &amp;&amp; !clientIdParam.equals(clientId)) {
+ *     return failure("client_id parameter does not match sub claim");
+ * }
+ *
+ * // UDS (logs and continues):
+ * if (clientIdParam != null &amp;&amp; !clientIdParam.equals(clientId)) {
+ *     LOGGER.debugf("client_id parameter '%s' does not match JWT subject '%s', ignoring", ...);
+ * }
+ * </pre>
+ *
+ * All other validation steps (assertion parameters, issuer, signature, audience, expiry,
+ * reuse detection) are unchanged from the upstream implementation.
+ *
+ * @see com.defenseunicorns.uds.keycloak.plugin.broker.kubernetes.UDSKubernetesIdentityProvider
+ */
+public class UDSFederatedJWTClientValidator extends FederatedJWTClientValidator {
+
+    private static final Logger LOGGER = Logger.getLogger(UDSFederatedJWTClientValidator.class);
+
+    public UDSFederatedJWTClientValidator(ClientAuthenticationFlowContext context,
+            SignatureValidator signatureValidator, String expectedTokenIssuer,
+            int allowedClockSkew, boolean reusePermitted) throws Exception {
+        super(context, signatureValidator, expectedTokenIssuer, allowedClockSkew, reusePermitted);
+    }
+
+    @Override
+    public boolean validate() {
+        return validateClientAssertionParametersInternal() &&
+                validateClientInternal() &&
+                validateSignatureAlgorithm(getExpectedSignatureAlgorithm()) &&
+                validateSignatureInternal() &&
+                validateTokenAudience(getExpectedAudiences(), isMultipleAudienceAllowed()) &&
+                validateTokenActive(getAllowedClockSkew(), getMaximumExpirationTime(), isReusePermitted());
+    }
+
+    /**
+     * Copied from {@code AbstractJWTClientValidator.validateClientAssertionParameters()} (private).
+     */
+    private boolean validateClientAssertionParametersInternal() {
+        String clientAssertionType = clientAssertionState.getClientAssertionType();
+        String clientAssertion = clientAssertionState.getClientAssertion();
+
+        if (clientAssertionType == null) {
+            return failure("Parameter client_assertion_type is missing");
+        }
+
+        if (!expectedClientAssertionType.equals(clientAssertionType)) {
+            return failure("Parameter client_assertion_type has value '"
+                    + clientAssertionType + "' but expected is '" + expectedClientAssertionType + "'");
+        }
+
+        if (clientAssertion == null) {
+            return failure("client_assertion parameter missing");
+        }
+
+        return true;
+    }
+
+    /**
+     * Modified from {@code AbstractJWTClientValidator.validateClient()} (private).
+     *
+     * <p>The only change: when {@code client_id} form parameter doesn't match the JWT subject,
+     * this logs a debug message instead of returning a failure. The {@code client_id} parameter
+     * is optional in the {@code client_credentials} grant with JWT assertion per RFC 7523.
+     */
+    private boolean validateClientInternal() {
+        JsonWebToken token = clientAssertionState.getToken();
+
+        String clientId = token.getSubject();
+        if (clientId == null) {
+            return failure("Token sub claim is required");
+        }
+
+        // Relaxed check: log instead of fail when client_id doesn't match JWT subject.
+        // In the K8s SA flow, client_id is "uds-operator" but sub is the SA identity.
+        String clientIdParam = context.getHttpRequest().getDecodedFormParameters().getFirst(OAuth2Constants.CLIENT_ID);
+        if (clientIdParam != null && !clientIdParam.equals(clientId)) {
+            LOGGER.debugf("client_id parameter '%s' does not match JWT subject '%s', ignoring mismatch",
+                clientIdParam, clientId);
+        }
+
+        String expectedTokenIssuer = getExpectedTokenIssuer();
+        if (expectedTokenIssuer != null && !expectedTokenIssuer.equals(token.getIssuer())) {
+            LOGGER.debugf("Expected token issuer '%s' does not match actual issuer '%s'",
+                expectedTokenIssuer, token.getIssuer());
+            return false;
+        }
+
+        ClientModel client = clientAssertionState.getClient();
+
+        if (client == null) {
+            context.failure(AuthenticationFlowError.CLIENT_NOT_FOUND, null);
+            return false;
+        } else {
+            context.getEvent().client(client.getClientId());
+            context.setClient(client);
+        }
+
+        if (!client.isEnabled()) {
+            context.failure(AuthenticationFlowError.CLIENT_DISABLED, null);
+            return false;
+        }
+
+        if (clientAuthenticatorProviderId != null && !clientAuthenticatorProviderId.equals(client.getClientAuthenticatorType())) {
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * Copied from {@code AbstractJWTClientValidator.validateSignature()} (private).
+     */
+    private boolean validateSignatureInternal() {
+        return signatureValidator.verifySignature(this);
+    }
+}
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Defense Unicorns
+ * Copyright 2026 Defense Unicorns
  * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
  */
 
@@ -107,17 +107,17 @@ public class UDSFederatedJWTClientValidator extends FederatedJWTClientValidator 
     private boolean validateClientInternal() {
         JsonWebToken token = clientAssertionState.getToken();
 
-        String clientId = token.getSubject();
-        if (clientId == null) {
+        String jwtSubject = token.getSubject();
+        if (jwtSubject == null) {
             return failure("Token sub claim is required");
         }
 
         // Relaxed check: log instead of fail when client_id doesn't match JWT subject.
         // In the K8s SA flow, client_id is "uds-operator" but sub is the SA identity.
         String clientIdParam = context.getHttpRequest().getDecodedFormParameters().getFirst(OAuth2Constants.CLIENT_ID);
-        if (clientIdParam != null && !clientIdParam.equals(clientId)) {
+        if (clientIdParam != null && !clientIdParam.equals(jwtSubject)) {
             LOGGER.debugf("client_id parameter '%s' does not match JWT subject '%s', ignoring mismatch",
-                clientIdParam, clientId);
+                clientIdParam, jwtSubject);
         }
 
         String expectedTokenIssuer = getExpectedTokenIssuer();
--- a/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/KubernetesUtils.java
+++ b/src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/KubernetesUtils.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2025 Defense Unicorns
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+ */
+
+package com.defenseunicorns.uds.keycloak.plugin.broker.kubernetes;
+
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+import org.jboss.logging.Logger;
+
+import static org.keycloak.broker.kubernetes.KubernetesConstants.SERVICE_ACCOUNT_TOKEN_PATH;
+
+/**
+ * Shared utilities for Kubernetes identity provider classes.
+ */
+final class KubernetesUtils {
+
+    private static final Logger LOGGER = Logger.getLogger(KubernetesUtils.class);
+
+    private KubernetesUtils() {
+    }
+
+    /**
+     * Reads the Kubernetes service account token from the standard mounted path.
+     *
+     * <p>On managed Kubernetes clusters (EKS, AKS, GKE), the SA token must be included in
+     * requests to the K8s API server's OIDC discovery and JWKS endpoints. The upstream Keycloak
+     * code conditionally includes this token based on an issuer match, which fails on managed
+     * clusters. Our custom classes use this method to always include the token.
+     *
+     * @return the SA token string, or {@code null} if the file doesn't exist or can't be read
+     */
+    static String readServiceAccountToken() {
+        try {
+            var path = Paths.get(SERVICE_ACCOUNT_TOKEN_PATH);
+            if (!Files.exists(path)) {
+                return null;
+            }
+            return new String(Files.readAllBytes(path), StandardCharsets.UTF_8).trim();
+        } catch (Exception e) {
+            LOGGER.warn("Failed to read service account token", e);
+            return null;
+        }
+    }
+}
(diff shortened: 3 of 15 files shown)


.ai/plans/signed_jwt_fix.md
```diff
@@ -0,0 +1,458 @@
+# Plan: Fix Signed JWT Authentication for Managed Kubernetes Clusters
```
slaskawi: This file will be removed once the testing is complete.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
```diff
@@ -0,0 +1,122 @@
...
+import org.keycloak.authentication.authenticators.client.FederatedJWTClientValidator;
+import org.keycloak.models.ClientModel;
+import org.keycloak.representations.JsonWebToken;
+
+import org.jboss.logging.Logger;
+
+/**
+ * Custom JWT client validator that relaxes the client_id form parameter check.
+ *
+ * This is a stop-gap solution until keycloak/keycloak#48026 is merged upstream.
```
slaskawi: Use the full issue link here.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
```diff
@@ -0,0 +1,146 @@
...
+            }
+
+            LOGGER.debugf("Discovered OIDC issuer '%s' for Kubernetes API server '%s'", discoveredIssuer, kubeApiServerUrl);
+            ISSUER_CACHE.put(kubeApiServerUrl, discoveredIssuer);
+            return discoveredIssuer;
+        } catch (Exception e) {
+            LOGGER.errorf(e, "Failed to discover OIDC issuer from '%s/.well-known/openid-configuration', falling back to configured issuer '%s'",
+                kubeApiServerUrl, kubeApiServerUrl);
+            return kubeApiServerUrl;
+        }
```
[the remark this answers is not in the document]
slaskawi: I think it's beneficial to leave it like this. It needs to fail very laudly as there's a terrible misconfiguration somewhere if this discovery doesn't work

tasks.yaml
```diff
@@ -130,7 +130,7 @@ tasks:
     description: "Clone UDS Core for integration testing"
     actions:
       - cmd: rm -rf uds-core
-      - cmd: git clone --branch main https://github.com/defenseunicorns/uds-core.git
+      - cmd: git clone --branch sebastianlaskawiec/core-152-use-signed-jwts-for-uds-operator-2 https://github.com/defenseunicorns/uds-core.git
```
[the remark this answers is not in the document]
slaskawi: Yes, This is necessary and may even be necessary for the patch release.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesJwksEndpointLoader.java
```diff
@@ -0,0 +1,88 @@
...
+
+        SimpleHttpRequest wellKnownRequest = simpleHttp.doGet(wellKnownEndpoint).acceptJson();
+        if (token != null) {
+            wellKnownRequest.auth(token);
+        }
+        String jwksUri = wellKnownRequest.asJson(OIDCConfigurationRepresentation.class).getJwksUri();
+
+        SimpleHttpRequest jwksRequest = simpleHttp.doGet(jwksUri).header(HttpHeaders.ACCEPT, "application/jwk-set+json");
+        if (token != null) {
+            jwksRequest.auth(token);
```
slaskawi, line 95: ## SSRF + Credential Leakage via Unvalidated `jwksUri`

The `jwksUri` is extracted from the OIDC discovery response and used directly as a URL with **no validation**. A compromised or MITM'd K8s API server could return a `jwksUri` pointing to `http://169.254.169.254/latest/meta-data/` (AWS IMDS), an internal service, or an attacker-controlled host. The SA token is then **forwarded to that URL** via `jwksRequest.auth(token)`, leaking a Kubernetes bearer credential.

The upstream code's issuer-match check (which this PR removes) was partially guarding against this. This code removes that guard entirely.

### Suggested Fix
- Validate `jwksUri` starts with `https://`
- Validate `jwksUri` hostname matches the configured K8s API server hostname (or is on an allowlist)
- Only attach the SA token when the `jwksUri` hostname matches the configured issuer hostname

---
*Found by the Brutal Code Review Skill*

slaskawi: Actually, this is false positive. We send the token to the issuer configured in the IdP. In other words, it's always https://kubernetes.default.svc.cluster.local, see https://github.com/defenseunicorns/uds-identity-config/pull/838/changes#diff-2c87f717454b99ca92e8c5b01951e81047d9f7e40268d45660b6a8e84a5cdcb6R1947

src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategyTest.java
```diff
@@ -0,0 +1,165 @@
...
+        // AKS issuer -- default lookup returns null (no matching IdP)
+        when(token.getIssuer()).thenReturn(AKS_ISSUER);
+        when(lookupProvider.lookupIdentityProviderFromIssuer(session, AKS_ISSUER)).thenReturn(null);
+
+        // Fallback: client has matching jwt.credential.sub, and jwt.credential.issuer points to uds-kubernetes IdP
+        when(clientModel.getAttribute("jwt.credential.sub")).thenReturn(SUBJECT);
+        when(clientModel.getAttribute("jwt.credential.issuer")).thenReturn("kubernetes");
+        when(clientModel.getClientId()).thenReturn("uds-operator");
+        when(realm.getClientsStream(anyInt(), anyInt())).thenReturn(Stream.of(clientModel));
+        when(idpStorageProvider.getByAlias("kubernetes")).thenReturn(udsIdpModel);
```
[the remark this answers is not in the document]
slaskawi: Fixed

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
```diff
@@ -0,0 +1,212 @@
...
+        String cached = ISSUER_CACHE.get(kubeApiServerUrl);
+        if (cached != null) {
+            LOGGER.debugf("Using cached OIDC issuer '%s' for K8s API server '%s'", cached, kubeApiServerUrl);
+            return cached;
+        }
+
+        String discovered = fetchIssuerFromOidc(kubeApiServerUrl);
+        // putIfAbsent is atomic -- if another thread raced us, use their result
+        String existing = ISSUER_CACHE.putIfAbsent(kubeApiServerUrl, discovered);
+        return existing != null ? existing : discovered;
```
chance-coleman, line 143: `fetchIssuerFromOidc` returns `kubeApiServerUrl` on any failure (connection error, 401, empty issuer, non-HTTPS issuer). That fallback value then gets stored in `ISSUER_CACHE` via `putIfAbsent`. I think if the K8s API server is briefly unavailable during the first auth attempt on EKS/AKS/GKE, the wrong issuer gets cached permanently until Keycloak restarts. Would it make sense to only cache when discovery actually succeeds?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesJwksEndpointLoader.java
```diff
@@ -0,0 +1,95 @@
...
+ * <h2>What this loader changes</h2>
+ *
+ * The only difference from upstream is that {@link KubernetesUtils#readServiceAccountToken()}
+ * always returns the SA token if the file exists, without checking the issuer claim. The SA
+ * token is safe to include unconditionally because it's already scoped to the cluster and is
+ * the same token the upstream code would include if the issuer matched.
+ *
+ * @see UDSKubernetesIdentityProvider
+ */
+public class UDSKubernetesJwksEndpointLoader implements PublicKeyLoader {
```
chance-coleman, line 55: Might be good to add some tests for this new file:
- happy path
- null jwks_uri

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
```diff
@@ -0,0 +1,212 @@
...
+
+    private String fetchIssuerFromOidc(String kubeApiServerUrl) {
+        String wellKnownEndpoint = kubeApiServerUrl + "/.well-known/openid-configuration";
+        try {
+            String saToken = KubernetesUtils.readServiceAccountToken();
+
+            SimpleHttp simpleHttp = SimpleHttp.create(session);
+            var request = simpleHttp.doGet(wellKnownEndpoint).acceptJson();
+            if (saToken != null) {
+                request.header("Authorization", "Bearer " + saToken);
```
chance-coleman: Inconsistent SA token injection: `fetchIssuerFromOidc` uses `request.header("Authorization", "Bearer " + saToken)` while `loadKeys()` from [here](https://github.com/defenseunicorns/uds-identity-config/pull/838/changes#diff-f2a3f541771aff3712f22d8ea9415356949a89601e6093c8087b5e11a5cba7fbR76)  uses `request.auth(token)` for the same endpoint. Does it make sense to align these?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
```diff
@@ -0,0 +1,113 @@
...
+                if (idp != null
+                        && UDSKubernetesIdentityProviderFactory.PROVIDER_ID.equals(idp.getProviderId())
+                        && idp.isEnabled()) {
+                    LOGGER.debugf("Fallback lookup matched client '%s' to IdP '%s'",
+                        matchingClient.getClientId(), idpAlias);
+                    return new LookupResult(matchingClient, idp);
+                }
+                return null;
+            })
+            .filter(r -> r != null)
```
chance-coleman: nit:

```suggestion
            .filter(Objects::nonNull)
```

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright 2025 Defense Unicorns
```
chance-coleman: nit: most files are dated as 2025

```suggestion
 * Copyright 2026 Defense Unicorns
```

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesJwksEndpointLoader.java
```diff
@@ -0,0 +1,95 @@
...
+        if (token != null) {
+            wellKnownRequest.auth(token);
+        }
+        String jwksUri = wellKnownRequest.asJson(OIDCConfigurationRepresentation.class).getJwksUri();
+        if (jwksUri == null || jwksUri.isEmpty()) {
+            throw new IllegalStateException("OIDC discovery at " + wellKnownEndpoint + " returned no jwks_uri");
+        }
+
+        // Forwarding a token along with the request is safe here. The Kubernetes API Server issuer is configured at the
+        // bootstrap time with the `--oidc-*` flags. So a successful attack would require controling these flags
```
chance-coleman:
```suggestion
        // bootstrap time with the `--oidc-*` flags. So a successful attack would require controlling these flags
```

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
joelmccoy, line 1: Maybe a silly question here... Would upstream have the appetite for maybe baking some of this lookup behavior into the upstream kubernetes provider?? It feels weird/unideal that the upstream kubernetes provider doesn't work with cloud native kubernetes deployments... And I would imagine some of this logic could/should be included in upstream.  Do you think there would be an appetite for that or is that even possible @slaskawi ?

mjnagel: Definitely +1 to this, it seems like a lot that would be valuable upstream. I think that was already your intent @slaskawi  (to push this back upstream over time), so maybe okay to keep in our own plugin code for now...but curious your thoughts here.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
```diff
@@ -0,0 +1,113 @@
...
+        if (subject == null) {
+            LOGGER.debug("Fallback lookup skipped because client assertion subject is missing");
+            return null;
+        }
+        var idpStorage = context.getSession().identityProviders();
+
+        return context.getRealm().getClientsStream()
+            .filter(c -> subject.equals(c.getAttribute(FederatedJWTClientAuthenticator.JWT_CREDENTIAL_SUBJECT_KEY)))
+            .map(matchingClient -> {
+                String idpAlias = matchingClient.getAttribute(FederatedJWTClientAuthenticator.JWT_CREDENTIAL_ISSUER_KEY);
```
mjnagel: I guess curious on the overall intent/usage of this custom "uds-kubernetes" provider. Is this something we'd envision going beyond just our uds-operator (and maybe tofu) client? Reason I ask - this matching on attribute + findFirst feels like it could have a few potential problems if I'm following right:
- volume of clients in a large keycloak could make this scan/filter slower (_probably_ not a real risk for most of users, but at scale maybe?)
- if multiple clients set the same `jwt.credential.sub` attribute...this could be a problem in looking up the right one

Basically wondering if we could simplify this (or other portions of our code) if we narrowly focused on making this work for just the operator and/or tofu clients. Maybe somewhat related to the other comment on pushing some of this upstream - would anything be simpler if we focused on our use-case and held off on generic support until we upstream that?

mjnagel (commented):
Seems like there's a build issue in CI that will definitely need to be fixed. Overall following the general flow of this but did have a few comments to help understand direction and if this is the right approach.

mjnagel:
@codex review

tasks.yaml
```diff
@@ -130,7 +130,7 @@ tasks:
...
```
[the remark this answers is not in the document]
mjnagel: Intentional for now

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
slaskawi: I updated the logic and now fail the entire auth flow if the discovery doesn't succeed. This approach should produce more consistent results.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesJwksEndpointLoader.java
slaskawi: Added new tests to UDSKubernetesJwksEndpointLoaderTest:

  - testLoadKeysThrowsOnEmptyJwksUri — covers the jwksUri.isEmpty() branch (line 79), complementing the existing null test
  - testLoadKeysSkipsAuthWhenNoSaToken — verifies that when no SA token file exists, auth() is never called on either request (the if (token != null) branches at lines 75 and 88)

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
slaskawi: Aligned.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
slaskawi: Absolutely! Once we have everything working (which means these changes released and running properly in IAC in UDS Core), I'll start an upstream discussion and see what parts we can push upstream. I would love to push everything, so in a couple months we can wipe this code entirely. That would be my goal!

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
slaskawi: > I guess curious on the overall intent/usage of this custom "uds-kubernetes" provider.

The UDS Kubenernetes IdP is a stop-gap solution we need for now to ensure the approach of scanning the Kubernetes API OIDC Discovery and following the `issuer` field is the correct approach. Unfortunately, this issue manifests itself only with managed Kubernetes clusters and there's a strong push back on integrating these in the Keycloak testsuite (which is why we didn't spot this earlier - we were testing on k3d and Minikube). 

> Is this something we'd envision going beyond just our uds-operator (and maybe tofu) client?

Yes! Since we already had to implement our custom IdP, I also relaxed the Client ID containstraint (similarly to https://github.com/keycloak/keycloak/pull/48026) to unblock the Tofu integration ([CORE-153](https://linear.app/defense-unicorns/issue/CORE-153/use-signed-jwts-for-the-tofu-client)). 

> Reason I ask - this matching on attribute + findFirst feels like it could have a few potential problems if I'm following right:

I _think_ the logic is correct. The configuration is divided into two parts:

* The IdP has `issuer` configured, which points to the Kube API Server.
* The Client has two attributes:
  * `jwt.credential.sub` that binds to a Service Accout, e.g. `system:serviceaccount:pepr-system:pepr-uds-core`.
  * `jwt.credential.issuer` which is an alias to the IdP, e.g. `uds`. 
  
Now the logic is the following:

1. We perform the standard (just like upstream) lookup first:
    - We extract the `issuer` from the signed JWT Token (from the `issuer` field) and try to find an IdP based on that.
    - Then we search for the Client based on `sub` claim from the JWT token and an IdP alias. 
    - In case when Kube is the issue (so k3d, minikube etc) - that's an exact match and we return here.
2. If the standard lookup fails, we try our own lookup:
    -  We find all Clients that have matching the `jwt.credential.sub` configurations. We expect this to be 1 already (in theory you can have many but in this case, we assume they are attached to different IdPs)
    - Then, based on that Client we search for an IdP following the alias. 
    - Finally, after a non-null check, we return what we found. Note that it is strongly expected to be 1.

The biggest difference between the two lookup methods is that first we find a proper Client and then search from the IdP. So we rely more on proper Keycloak Client configuration rather than the JWT token (as we know that the `issuer` won't match). 

I think this could be refined further but the implementation aims to be as relaxed as possible. When pushing things to the upstream we'll probably tighten it up.

To answer your direct question about Tofu - this should work quite well in that case too. Note, the Tofu will have its own Client assigned to its own Service Account (not Pepr Service Account). If I haven't missed anything, during the second lookup we should quickly select proper Client based on the Service Account (based on the `sub` claim, which probably will be something like `system:serviceaccount:keycloak:open-tofu-client`). 

> Basically wondering if we could simplify this (or other portions of our code) if we narrowly focused on making this work for just the operator and/or tofu clients. Maybe somewhat related to the other comment on pushing some of this upstream - would anything be simpler if we focused on our use-case and held off on generic support until we upstream that?

I think it will be hard to come up with something even simpler without putting strong assumptiong (like names) into the picture. Note, that we already have one very strongly limiting condition there to avoid selecting things we don't want: 

```
                if (idp != null
                        && UDSKubernetesIdentityProviderFactory.PROVIDER_ID.equals(idp.getProviderId())
                        && idp.isEnabled()) {
```

This ensures the second lookup selects only our own IdPs making it a match (because Client Assertions are designed to match two things at once - Clients and IdPs).

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
```diff
@@ -0,0 +1,212 @@
...
+    String discoverIssuer(String kubeApiServerUrl) {
+        String cached = ISSUER_CACHE.get(kubeApiServerUrl);
+        if (cached != null) {
+            LOGGER.debugf("Using cached OIDC issuer '%s' for K8s API server '%s'", cached, kubeApiServerUrl);
+            return cached;
+        }
+
+        String discovered = fetchIssuerFromOidc(kubeApiServerUrl);
+        // putIfAbsent is atomic -- if another thread raced us, use their result
+        String existing = ISSUER_CACHE.putIfAbsent(kubeApiServerUrl, discovered);
```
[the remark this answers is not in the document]
slaskawi: I think it's a valid concern. Addressed along with @chance-coleman 's comment.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
slaskawi: I started gathering feedback here: https://github.com/keycloak/keycloak/discussions/48130

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
mjnagel: :+1: good with this, mostly just wanted to ask if it would make sense to restrict in our case to the two client names (tofu/operator). Following the general approach and where this comes into the picture. I'll resolve this for now.

Edit: after further review going to make a few specific comments on the places I'm thinking we could change slightly to improve handling/efficiency.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesJwksEndpointLoader.java
mjnagel: This came up in my local review as well and I think it is a valid finding? `jwksUri` in this case is coming from `wellKnownRequest`. So _yes_ we always hit `https://kubernetes.default.svc.cluster.local/.well-known/openid-configuration` first here, but then we are taking the `jwksUri` from that response (in EKS something like `https://oidc.eks.us-east-1.amazonaws.com...`) which is where our token ultimately gets forwarded to.

I think this can be a pretty basic fix to just not forward the token when hitting a different jwksUri, with 0 downside? Something like:
```java
  if (token != null && isSameHost(jwksUri, issuer)) {                                                                                                                                                                                        
      jwksRequest.auth(token);
  }
```

Where `isSameHost` is a basic helper to make sure the hosts are equal.

Basically I think you're correct about the _issuer_ request, but the _jwks_ request could actually be a different endpoint where we don't need/want to send the SA token.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
```diff
@@ -0,0 +1,115 @@
...
+                        && UDSKubernetesIdentityProviderFactory.PROVIDER_ID.equals(idp.getProviderId())
+                        && idp.isEnabled()) {
+                    LOGGER.debugf("Fallback lookup matched client '%s' to IdP '%s'",
+                        matchingClient.getClientId(), idpAlias);
+                    return new LookupResult(matchingClient, idp);
+                }
+                return null;
+            })
+            .filter(Objects::nonNull)
+            .findFirst()
```
mjnagel: On second thought 🫣 related to my previous comment - I wonder if it's better to clearly fail closed here if we end up with multiple matches? Right now we just return the first match, but it might make sense to check for `> 1` match and refuse to handle an ambiguous auth path (or some other handling)?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesJwksEndpointLoader.java
```diff
@@ -0,0 +1,95 @@
...
+        SimpleHttp simpleHttp = SimpleHttp.create(session);
+        String token = KubernetesUtils.readServiceAccountToken();
+
+        String wellKnownEndpoint = issuer + "/.well-known/openid-configuration";
+
+        SimpleHttpRequest wellKnownRequest = simpleHttp.doGet(wellKnownEndpoint).acceptJson();
+        if (token != null) {
+            wellKnownRequest.auth(token);
+        }
+        String jwksUri = wellKnownRequest.asJson(OIDCConfigurationRepresentation.class).getJwksUri();
```
mjnagel: Should we check the http status here before trying to deserialize? Similar to what we do in `fetchIssuerFromOidc`.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
```diff
@@ -0,0 +1,115 @@
...
+        ClientAssertionState state = context.getState(ClientAssertionState.class, ClientAssertionState.supplier());
+        String subject = state != null && state.getToken() != null ? state.getToken().getSubject() : null;
+        if (subject == null) {
+            LOGGER.debug("Fallback lookup skipped because client assertion subject is missing");
+            return null;
+        }
+        var idpStorage = context.getSession().identityProviders();
+
+        return context.getRealm().getClientsStream()
+            .filter(c -> subject.equals(c.getAttribute(FederatedJWTClientAuthenticator.JWT_CREDENTIAL_SUBJECT_KEY)))
```
mjnagel: Could we use [`ClientProvider.searchClientsByAttributes()`](https://www.keycloak.org/docs-api/latest/javadocs/org/keycloak/storage/client/ClientLookupProvider.html#searchClientsByAttributes(org.keycloak.models.RealmModel,java.util.Map,java.lang.Integer,java.lang.Integer)) here instead, which might be slightly more efficient than grabbing all clients and filtering ourselves?

mjnagel (commented):
Few follow-on comments. No major concerns but think there's a few places to consider improving handling/logic.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesJwksEndpointLoader.java
chance-coleman: Both this line and line 92 deserialize with no status check. I think a failed call here would result in a parse exception that has no context about the failed URL or status.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
```diff
@@ -0,0 +1,158 @@
...
+            LOGGER.debugf("client_id parameter '%s' does not match JWT subject '%s', ignoring mismatch",
+                clientIdParam, clientId);
+        }
+
+        String expectedTokenIssuer = getExpectedTokenIssuer();
+        if (expectedTokenIssuer != null && !expectedTokenIssuer.equals(token.getIssuer())) {
+            LOGGER.debugf("Expected token issuer '%s' does not match actual issuer '%s'",
+                expectedTokenIssuer, token.getIssuer());
+            return false;
+        }
```
chance-coleman, line 128: Every other false path in `validateClientInternal()` calls `context.failure(AuthenticationFlowError.*)`, should we also call that here? If i'm understanding this specific mismatch case it would be security event that we would want audit events for.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
```diff
@@ -0,0 +1,234 @@
...
+    public boolean verifyClientAssertion(ClientAuthenticationFlowContext context) throws Exception {
+        String discoveredIssuer = discoverIssuer(config.getIssuer());
+        if (discoveredIssuer == null) {
+            Response challengeResponse = ClientAuthUtil.errorResponse(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), "server_error", "Failed to discover OIDC issuer from Kubernetes API server");
+            context.failure(AuthenticationFlowError.INTERNAL_ERROR, challengeResponse);
+            return false;
+        }
+
+        UDSFederatedJWTClientValidator validator = new UDSFederatedJWTClientValidator(
+            context, this::verifySignature, discoveredIssuer, config.getAllowedClockSkew(), true
```
chance-coleman, line 109: for my own understanding, is there ever a time that reuse would not be permitted? Just wondering about hard coding that to `true` here.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
```diff
@@ -0,0 +1,234 @@
...
+        );
+        validator.setMaximumExpirationTime(3600);
```
chance-coleman, line 111: in a similar vein to previous comment, would we ever have a usecase where an admin would want to have a different expiration timeout?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
```diff
@@ -0,0 +1,234 @@
...
+     *
+     * <p>On k3d/self-hosted clusters: returns the same configured issuer (no-op).
+     * On EKS/AKS/GKE: returns the managed cluster's public OIDC issuer URL.
+     *
+     * <p>The SA token is included in the request because managed K8s API servers (notably EKS)
+     * return 401 Unauthorized for unauthenticated requests to the OIDC discovery endpoint
+     * when accessed via the internal {@code kubernetes.default.svc.cluster.local} URL.
+     *
+     * @param kubeApiServerUrl the configured K8s API server URL (e.g. https://kubernetes.default.svc.cluster.local)
+     * @return the discovered issuer URL, or {@code kubeApiServerUrl} as fallback on any failure
```
chance-coleman: nit: i think this javadoc is wrong, it returns null on failure now.
```suggestion
     * @return the discovered issuer URL, or {@code null} if OIDC discovery fails
```

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
```diff
@@ -0,0 +1,158 @@
...
+     * Modified from {@code AbstractJWTClientValidator.validateClient()} (private).
+     *
+     * <p>The only change: when {@code client_id} form parameter doesn't match the JWT subject,
+     * this logs a debug message instead of returning a failure. The {@code client_id} parameter
+     * is optional in the {@code client_credentials} grant with JWT assertion per RFC 7523.
+     */
+    private boolean validateClientInternal() {
+        JsonWebToken token = clientAssertionState.getToken();
+
+        String clientId = token.getSubject();
```
chance-coleman: nit: based on the javadoc this class is meant to handle the `sub` is **not** the `client_id`. I think it might make more sense to name this variable `jwtSubject`.

```suggestion
        String jwtSubject = token.getSubject();
```

chance-coleman (commented):
few more comments and a couple questions.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
```diff
@@ -0,0 +1,158 @@
+/*
+ * Copyright 2025 Defense Unicorns
```
joelmccoy: nit: there are alot of files generated with the 2025 instead of the 2026 copyright date

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * Custom JWT client validator that relaxes the {@code client_id} form parameter check.
+ *
+ * <h2>Why this exists</h2>
+ *
+ * The upstream {@link FederatedJWTClientValidator} inherits from {@code AbstractJWTClientValidator}
+ * which has a {@code private} method {@code validateClient()} that rejects requests where the
+ * {@code client_id} form parameter doesn't match the JWT {@code sub} claim. In the Kubernetes
+ * service account flow, the JWT subject is the SA identity
+ * (e.g. {@code system:serviceaccount:pepr-system:pepr-uds-core}) while the {@code client_id}
```
joelmccoy, line 26: Maybe this is another silly question, but is there a strong reason we shouldn't create a new client in Keycloak and name it the service account we expect Pepr to have, `system:serviceaccount:pepr-system:pepr-uds-core` instead of taking this approach?

Relaxing the `client_id` check feels counterintuitive, and I'm unsure about the security implications of doing this.

If two Keycloak clients have the same `jwt.credential.sub` value, `UDSClientAssertionStrategy` returns the first one it finds with `.findFirst()`. A misconfiguration could map a service account to the wrong client, which the upstream `client_id` check would have caught?? But maybe if we remove the `findFirst()` and throw an error instead maybe this isn't concern.

Don't want to derail this whole approach, but I thought this was a question worth asking.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
```diff
@@ -0,0 +1,158 @@
...
+ * parameter is the Keycloak client name (e.g. {@code uds-operator}). These intentionally differ.
+ *
+ * <p>Because {@code validateClient()} is private, we cannot override just that method. We must
+ * override {@code validate()} entirely and copy the private helper methods, modifying only the
+ * client_id check to log a debug message instead of failing.
+ *
+ * <p>This is a stop-gap solution until
+ * <a href="https://github.com/keycloak/keycloak/pull/48026">https://github.com/keycloak/keycloak/pull/48026</a>
```
joelmccoy, line 34: If we stick with this approach, can we create a linear ticket to clean this up and remove this logic? Currently, we are copying upstream functions directly, and we definitely want to avoid any drift if possible. We don't want to lose track.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
joelmccoy: I had this same thought... But it looks like it matches upstream exactly? https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/authentication/authenticators/client/AbstractJWTClientValidator.java#L111-L113

Probably good to keep in line with upstream? If this is an issue, probably best to fix upstream

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/KubernetesUtils.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(KubernetesUtils.class);
+
+    private KubernetesUtils() {
+    }
+
+    /**
+     * Reads the Kubernetes service account token from the standard mounted path.
+     *
+     * <p>On managed Kubernetes clusters (EKS, AKS, GKE), the SA token must be included in
```
joelmccoy: nit: can we reference the upstream code reference for this like we do for other functions? it is my understanding that this function is basically the same as a upstream one but doesn't verify the issuer?  if my understanding is incorrect, you can ignore this.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/KubernetesUtils.java
```diff
@@ -0,0 +1,48 @@
...
+     * requests to the K8s API server's OIDC discovery and JWKS endpoints. The upstream Keycloak
+     * code conditionally includes this token based on an issuer match, which fails on managed
+     * clusters. Our custom classes use this method to always include the token.
+     *
+     * @return the SA token string, or {@code null} if the file doesn't exist or can't be read
+     */
+    static String readServiceAccountToken() {
```
joelmccoy, line 72: nit: maybe add some basic tests for this

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesJwksEndpointLoader.java
joelmccoy: this came back as finding for me too. would recommend we do this

joelmccoy (commented):
I believe this approach is effective, but I have some questions and concerns to discuss with the group:

1. Upstream code copying: We are copying several private methods from `AbstractJWTClientValidator` and `KubernetesIdentityProvider` because they cannot be overridden. These copies will lock in the behavior of Keycloak 26.5 and may silently drift during upgrades. [keycloak/keycloak#48026 ](https://github.com/keycloak/keycloak/pull/48026)would allow for extensible validation potentially?. Is it worth waiting for that, or do we need signed JWTs now enough to accept the maintenance burden?

2. Relaxed `client_id` validation: The custom validator skips the upstream check that the `client_id` form parameter matches the JWT `sub`. Could we avoid this entirely by renaming/creating a new Keycloak client to the service account identity (`system:serviceaccount:pepr-system:pepr-uds-core`)? This would eliminate the need for the entire `UDSFederatedJWTClientValidator` class, reducing the number of copied upstream files to maintain. Not sure if this is even possible but wanted to ask?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesJwksEndpointLoader.java
slaskawi: Unfortunately it's not that easy. The `issuer` here comes from IdP's `config.getIssuer()`. This means it's always `https://kubernetes.default.svc.cluster.local`. The JWKS URL points to a centralized IdP, so there will always be a mismatch. 

But I think I found a better way. I'm extracting the `iss` from the SA Token and if it matches the JWKS origin, we know we're sending a token to a trusted domain. 

I'll test it out and see if it actually covers the EKS case (it should).

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
slaskawi: Added a check. It should report the root cause in the logs but send `invalid_credentials` to the Client.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesJwksEndpointLoader.java
slaskawi: Fixed but the solution isn't pretty. Keycloak doesn't have any good utility for checking if the response was successful so I had to create my own.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSClientAssertionStrategy.java
slaskawi: Of course! Fixed.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesJwksEndpointLoader.java
slaskawi: Yes, I've implemented this changes already. Will be available in the next patch.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
slaskawi: I try to keep the method signatures 1:1 with the upstream, which will make backporting easier. So even though you have a good point, let's keep it as is.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
slaskawi: Similarly to the above - 100% agree but let's keep the method signatures similar to the upstream.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProvider.java
slaskawi: Fixed.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
slaskawi: Done

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
slaskawi: Updated.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
slaskawi: >Maybe this is another silly question, but is there a strong reason we shouldn't create a new client in Keycloak and name it the service account we expect Pepr to have, system:serviceaccount:pepr-system:pepr-uds-core instead of taking this approach?

> Relaxing the client_id check feels counterintuitive, and I'm unsure about the security implications of doing this.

Client Assertions is an auth method that binds the issuer and a token together. Such a relationship in Keycloak is created by using IdPs. Going forward with this, the Client ID is not needed at all in this scenario. Keycloak needs to find proper IdP based on the token issuer and not Client ID. As mentioned https://github.com/keycloak/keycloak/discussions/48130, most of the libraries start to adapt and do not send Client ID when using Client Assertions. That's the correct approach. Unfortunately as I highlighted in https://github.com/keycloak/terraform-provider-keycloak/issues/1542, Keycloak Terraform Provided didn't manage to catch up so far (although I already issued a PRs for that, see https://github.com/keycloak/terraform-provider-keycloak/pull/1552 and https://github.com/keycloak/keycloak/pull/48026).

To sum it up - sending Client ID doesn't improve security in any measure and is more of a mistake when using Client Assertions. Relaxing the checks make perfect sense as Keycloak identifies the Client using different method than a hint sent by the Client app. 

> If two Keycloak clients have the same jwt.credential.sub value, UDSClientAssertionStrategy returns the first one it finds with .findFirst(). A misconfiguration could map a service account to the wrong client, which the upstream client_id check would have caught?? But maybe if we remove the findFirst() and throw an error instead maybe this isn't concern.

@mjnagel asked almost identical question here: https://github.com/defenseunicorns/uds-identity-config/pull/838#discussion_r3090408220

The code has been refactored in this area and it now fails laudly if there are more than 2 Client + IdP pairs found.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/authentication/authenticators/client/UDSFederatedJWTClientValidator.java
slaskawi: Done: https://linear.app/defense-unicorns/issue/CLI-150/remove-copied-upstream-methods-from-udsfederatedjwtclientvalidator

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/KubernetesUtils.java
slaskawi: Sure, added.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/KubernetesUtils.java
slaskawi: Done.

slaskawi:
> Upstream code copying: We are copying several private methods from AbstractJWTClientValidator and KubernetesIdentityProvider because they cannot be overridden. These copies will lock in the behavior of Keycloak 26.5 and may silently drift during upgrades. https://github.com/keycloak/keycloak/pull/48026 would allow for extensible validation potentially?. Is it worth waiting for that, or do we need signed JWTs now enough to accept the maintenance burden?

There's actually much more than that. The https://github.com/keycloak/keycloak/discussions/48130 summerizes most of my concerns and it will result in potentially a few follow-up tickets in Keycloak. We just need to agree on the direction.

This means we have two choices: we can press on with what we have or revert back to Client ID and Kubernetes Secret and wait until the upstream changes are ready. At the moment, Keycloak has a new minor version every 3 months so we're probably looking at 3-6 months time horizon (depending on how quickly we can reach consensus). 

> Relaxed client_id validation: The custom validator skips the upstream check that the client_id form parameter matches the JWT sub. Could we avoid this entirely by renaming/creating a new Keycloak client to the service account identity (system:serviceaccount:pepr-system:pepr-uds-core)? This would eliminate the need for the entire UDSFederatedJWTClientValidator class, reducing the number of copied upstream files to maintain. Not sure if this is even possible but wanted to ask?

Yes, that's correct but I don't think that's a sustaintable approach for the long run. As I explained in the previous comments - sending Client ID with Client Assertions is more of a mistake that we see at the early adoption stages.

slaskawi:
After a longer conversation, we decided to delay this feature until the Keycloak upstream is ready. 

Further follow-ups can be tracked within: https://github.com/keycloak/keycloak/discussions/48130

PR #841, opened 2026-04-20
feat(CORE-27): allow opt-in public clients with enforced PKCE

slaskawi:
## Summary

⚠️ This Pull Request depends on https://github.com/defenseunicorns/uds-core/pull/2598

This Pull Request introduces the `Public Clients` switch into the UDS Client Policy, which essentially makes the entire feature opt-in.

BEGIN_COMMIT_OVERRIDE
ci: test opt-in public clients with enforced PKCE (#841)
END_COMMIT_OVERRIDE


src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -232,10 +234,28 @@ void validateClientSettings(ClientRepresentation rep) throws ClientPolicyExcepti
...
+            return;
+        }
+
+        if (!configuration.isAllowPublicClients()) {
+            throw new ClientPolicyException(Errors.INVALID_CLIENT,
+                    "Public Clients are not allowed. Enable allow-public-clients on the UDS Client Profile to permit PKCE-protected public Clients.");
+        }
+
+        String pkceMethod = rep.getAttributes() == null ? null : rep.getAttributes().get(ATTRIBUTE_PKCE_CODE_CHALLENGE_METHOD);
+        if (!PKCE_METHOD_S256.equals(pkceMethod)) {
```
slaskawi: Double check if we can give Mission Heros more flexibility and tollerate any PKCE verification method (only only S256)

src/test/cypress/e2e/tofu-client.cy.ts
```diff
@@ -45,6 +45,7 @@ describe("Tofu Client Management", () => {
                 standard_flow_enabled        = true
                 direct_access_grants_enabled = true
                 service_accounts_enabled     = false
+                pkce_code_challenge_method   = "S256"
```
slaskawi: This shouldn't be needed

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -212,6 +237,12 @@ public void setupConfiguration(UDSClientPolicyPermissionsExecutorConfiguration c
             allowedClientScopes.addAll(config.getAllowedClientScopes());
         config.setAllowedClientScopes(new ArrayList<>(allowedClientScopes));
 
+        // Surface the effective allow-public-clients value at INFO so operators can audit the
+        // resolved boolean without enabling trace logs. The field is populated via Jackson
+        // deserialization from realm.json5, which comes from an env-var interpolation
+        // (${REALM_SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED:false}) — string coercion is
+        // brittle, and this log makes misconfiguration visible at startup.
+        logger.info("UDSClientPolicyPermissionsExecutor initialized: allow-public-clients={}", config.isAllowPublicClients());
         logger.debug("Initializing with configuration: {}", config);
```
slaskawi, line 236: This is wrong - ensure the `allowPublicClients` is visible in `toString` method and remove this (unnecessary) logging line.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -226,16 +257,181 @@ public String getProviderId() {
...
+     *       attribute to a non-blank value. Keycloak accepts {@code S256} (recommended)
+     *       and {@code plain} (discouraged) at the authorization endpoint per RFC 7636.
+     *       We accept any configured value and defer enforcement of the specific method
+     *       to Keycloak's runtime.</li>
+     * </ol>
+     * <p>
+     * <b>UPDATE semantics.</b> Keycloak's admin REST API accepts sparse UPDATE bodies:
+     * fields omitted from the proposed representation inherit their value from the
+     * stored Client. A policy check that looked only at the proposed rep would be
+     * trivially bypassable via a PATCH-like UPDATE (e.g. flipping {@code publicClient}
```
slaskawi: This seems to be totally wrong - Keycloak Admin API doesn't support HTTP PATCH commands. This means, we'll never receive delta object like this. This entire logic is unnecessary.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
slaskawi: Confirmed — the executor intentionally accepts any non-blank value for `pkce.code.challenge.method`. The UDS policy only proves PKCE is configured on the public Client; Keycloak validates the exact challenge method (`S256`, `plain`, etc.) at the authorization endpoint per RFC 7636.

Locked in by four unit tests in `UDSClientPolicyPermissionsExecutorTest` (commit c1d6ae2):
- `shouldAcceptPublicClientWithS256Pkce`
- `shouldAcceptPublicClientWithPlainPkce`
- `shouldAcceptPublicClientWithLowercaseS256Pkce`
- `shouldAcceptPublicClientWithArbitraryPkceMethodValue`
- `shouldAcceptPublicClientWithWhitespacePaddedPkceMethod`

Plus rejection tests for `null`, missing, and blank values. Javadoc on `validatePublicClient` spells out the decision.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -50,13 +52,33 @@
...
+    public static final String ATTRIBUTE_PKCE_CODE_CHALLENGE_METHOD = "pkce.code.challenge.method";
+
+    // Keycloak OIDC client attribute toggling RFC 8628 Device Authorization Grant. When
+    // true on a client with standardFlow/implicitFlow disabled, the client is effectively
+    // a device-flow-only public client. PKCE does not apply to device flow.
+    public static final String ATTRIBUTE_DEVICE_FLOW_ENABLED = "oauth2.device.authorization.grant.enabled";
+
+    public static final String PROTOCOL_OPENID_CONNECT = "openid-connect";
+
+    // Loopback hosts where plaintext http:// redirect URIs are permitted for public clients.
```
slaskawi, line 74: This doesn't belong to this piece - it's part of the Keycloak core validation

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
slaskawi: You're right. Verified in Keycloak 26.5.7: `services/src/main/java/org/keycloak/services/resources/admin/ClientResource.java` exposes only @PUT/@POST/@GET/@DELETE — no @PATCH. Every REGISTER and UPDATE carries a full `ClientRepresentation`, so the effective-merge helpers were solving a problem Keycloak doesn't have.

Removed in c61eb71:
- `validateClientSettings` / `validatePublicClient` no longer take a `ClientModel target`
- `effectivePublicClient`, `effectiveProtocol`, `effectiveBoolean`, `effectiveAttribute`, `effectiveRedirectUris` deleted
- `isDeviceFlowOnly` simplified to consult only the proposed rep
- UPDATE-path tests (`shouldRejectUpdateFlippingConfidentialClientToPublicViaSparseBody`, `shouldAcceptUpdateOnExistingCompliantPublicClientViaSparseBody`, `shouldRejectUpdateClearingPkceOnExistingPublicClient`) deleted
- Javadoc updated to record that the Admin API doesn't support PATCH so the proposed rep is the full post-change state

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -229,13 +256,136 @@ public String getProviderId() {
...
+     * direct uds-operator admin-API calls (integration tests, manual operator
+     * intervention) that bypass Pepr. It must remain a superset of Pepr's rules so
+     * bypassing Pepr never results in a weaker posture.
+     * <p>
+     * Keycloak's Admin API does not support HTTP PATCH on Clients — the ClientResource
+     * exposes only PUT/POST/GET/DELETE — so every REGISTER and UPDATE carries a full
+     * ClientRepresentation. We therefore validate the proposed rep directly, without
+     * merging against the stored {@link ClientModel}.
+     * <p>
+     * Enforcement order:
```
slaskawi, line 279: ⚠️ Note for reviewers:

Since Public Clients do not require any means of Client Authentication, they can be used as a stepping stone for an attack. In order to mitigate this (at least partially), Claude suggested implementing strictier Client Configuration validation. The biggest downside of this approach is that it's asymmetric comparing to Confidential Clients (e.g. you can set a wildcard redirect URL with Confidential Clients but not here). 

I'm slighly leaning towards accepting this lack of symmetry and keeping this piece. 

What do you think?

tasks.yaml
```diff
@@ -130,7 +130,7 @@ tasks:
     description: "Clone UDS Core for integration testing"
     actions:
       - cmd: rm -rf uds-core
-      - cmd: git clone --branch main https://github.com/defenseunicorns/uds-core.git
+      - cmd: git clone --branch sebastianlaskawiec/core-27 https://github.com/defenseunicorns/uds-core.git
```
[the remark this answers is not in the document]
slaskawi: This will be reverted once the UDS Core PR gets merged.

src/test/cypress/e2e/tofu-client.cy.ts
slaskawi: You're right — not needed. The UDS client-policy executor only runs for `uds-operator`-authenticated events (see `UDSClientPolicyPermissionsExecutor.java:134`: `UDS_OPERATOR_CLIENT_ID.equals(getAuthenticatedClientId(...))`). This test authenticates Terraform as `uds-opentofu-client`, a different principal, so the CORE-27 policy never runs for it.

Fully reverted `tofu-client.cy.ts` to origin/main in d811917 — zero diff vs main.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
slaskawi: Let's leave it for now - maybe during reviews we'll remove it.

slaskawi: Acknowledged — leaving the redirect-URI block as-is for this PR.

For the follow-up: Keycloak already ships a purpose-built `SecureRedirectUrisEnforcerExecutor` (services/src/main/java/org/keycloak/services/clientpolicy/executor/SecureRedirectUrisEnforcerExecutor.java, provider id `secure-redirect-uris-enforcer`) with knobs that map 1:1 to what `validatePublicClientRedirectUris` does here — `allow-http-scheme`, `allow-wildcards-in-redirect-uri`, `allow-ipv4-loopback-address`, `allow-ipv6-loopback-address`, `allow-permitted-domains`, `oauth-2-1-compliant`. When we revisit, the cleanest move is to delete `validatePublicClientRedirectUris` and wire that built-in executor into the UDS Client Profile in `realm.json5` instead.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
slaskawi: Done in f6cb0bd. Confirmed `allowPublicClients` is already surfaced via `Configuration.toString()` (line 101 of `UDSClientPolicyPermissionsExecutorConfiguration`), and `setupConfiguration` already logs the full config at DEBUG — the dedicated INFO line was duplicative. Removed.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutorFactory.java
```diff
@@ -74,6 +76,17 @@ public class UDSClientPolicyPermissionsExecutorFactory implements ClientPolicyEx
         useDefaultAllowedClientScopes.setType(ProviderConfigProperty.BOOLEAN_TYPE);
         useDefaultAllowedClientScopes.setDefaultValue(Boolean.TRUE.toString());
         configProperties.add(useDefaultAllowedClientScopes);
+
+        allowPublicClients.setName(ALLOW_PUBLIC_CLIENTS);
+        allowPublicClients.setLabel("Allow Public Clients");
+        allowPublicClients.setHelpText(
+                "When true, allows publicClient=true only when the client sets the " + UDSClientPolicyPermissionsExecutor.ATTRIBUTE_PKCE_CODE_CHALLENGE_METHOD
```
slaskawi, line 83: This comment should be shorter.

src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutorTest.java
```diff
@@ -149,56 +149,330 @@ public void shouldEnforceWhitelistedCustomClaims() {
...
+    }
+
+    @Test
+    public void shouldAcceptConfidentialClientWhenPublicFlagIsNull() throws ClientPolicyException {
+        ClientRepresentation rep = new ClientRepresentation();
+        // isPublicClient() == null — treat as confidential
+        executor.validateClientSettings(rep);
+    }
+
+    // ---- allow-public-clients=true: PKCE enforcement ----
```
slaskawi, line 182: Remove the unnecessary comment.

src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutorTest.java
```diff
@@ -149,56 +149,330 @@ public void shouldEnforceWhitelistedCustomClaims() {
...
+    }
+
+    @Test
+    public void shouldAcceptPublicClientWithWhitespacePaddedPkceMethod() throws ClientPolicyException {
+        // Whitespace-padded values still carry a method after trimming — accept them.
+        setupExecutor(true);
+        executor.validateClientSettings(publicClientWithPkce("  S256  "));
+    }
+
+    // ---- Protocol / flow / redirect-URI rules ----
```
slaskawi, line 248: Remove the unnecessary comment

src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutorTest.java
```diff
@@ -149,56 +149,330 @@ public void shouldEnforceWhitelistedCustomClaims() {
...
+
+    @Test
+    public void shouldAcceptPublicClientWithHttpsRedirectUri() throws ClientPolicyException {
+        setupExecutor(true);
+        ClientRepresentation rep = publicClientWithPkce("S256");
+        rep.setRedirectUris(new ArrayList<>(List.of("https://app.example.com/cb")));
+        executor.validateClientSettings(rep);
+    }
+
+    // ---- Device-flow-only exemption ----
```
slaskawi, line 357: Remove the unnecessary comment.

src/test/cypress/e2e/client-credentials.cy.ts
```diff
@@ -23,7 +23,10 @@ describe("UDS Operator Client Credentials", () => {
                     clientId: randomClientId,
                     enabled: true,
                     publicClient: true,
-                    redirectUris: ['http://localhost/*']
+                    redirectUris: ['http://localhost/*'],
+                    attributes: {
+                        'pkce.code.challenge.method': 'S256'
```
slaskawi, line 28: Is this required?

slaskawi: Yes — with the refactored plugin, the PKCE attribute is required here, but only because the test uses `publicClient: true`. Logic:

- `k3d-core-slim-dev` sets `SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED: "true"` → plugin `allow-public-clients=true`.
- The test client is not Device-Flow-only (no `oauth2.device.authorization.grant.enabled`, `standardFlowEnabled` defaults to `true`).
- It falls through to the non-Device-Flow hardening path and the PKCE check rejects it without `pkce.code.challenge.method`.

However, the test itself ("UDS Operator can delete its own Client") does not care whether the client is public or confidential — it just needs a client owned by `uds-operator` that it can then delete. The cleaner fix is to drop `publicClient: true`, which bypasses the public-client policy entirely and makes the PKCE attribute unnecessary. I will switch it to a confidential client in a follow-up commit.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -229,13 +251,185 @@ public String getProviderId() {
...
 
+    /**
+     * Enforces UDS hardening rules for {@code publicClient=true} Clients.
+     * <p>
+     * This executor is a companion to the Pepr admission validator in uds-core
+     * ({@code src/pepr/operator/crd/validators/package-validator.ts}). Pepr rejects
+     * non-compliant Clients declared on Package CRs; this executor catches direct
+     * uds-operator Admin-API calls (integration tests, manual operator intervention)
+     * that bypass Pepr. It must remain a superset of Pepr's rules so bypassing Pepr
+     * never results in a weaker posture.
```
bburky: I generally like doing this in as policy in Keycloak, but if current UDS cluster admins have already been going around the UDS operator and creating public clients in Keycloak, this is a breaking change to begin enforcing. It's probably fine though? We just need to document clearly that they need to enable `SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED` _and_ manually turn on the policy in Keycloak (variable only applies at initial realm import, existing clusters need manual configuration).

Make sure we document this very clearly

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -229,13 +251,185 @@ public String getProviderId() {
...
+                throw new ClientPolicyException(Errors.INVALID_CLIENT,
+                        "The Client is a public Client with a blank redirectUri. Rejecting request.");
+            }
+            // Keycloak treats a leading '*' as a host-level wildcard match. Combined with
+            // the inability to authenticate a public client during code exchange, this is
+            // a classic redirect-URI-injection foothold.
+            if (uri.equals("*") || uri.startsWith("*")) {
+                throw new ClientPolicyException(Errors.INVALID_CLIENT,
+                        "The Client is a public Client with a wildcard redirectUri " + uri + ". Rejecting request.");
+            }
```
bburky: What is this defending against? Accidentally overbroad wildcards? I guess it makes sense to block those. Or blocking a bare `*` matching any top level domain?

But it's not impossible that an app actually uses multiple subdomains and may really want `*.app.example.com`.

I'm fine with however, I just want to understand the intent of this check.

Also for the current code, `uri.equals("*") || ` is unneeded, `uri.startsWith("*")` is a superset of that check.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
mjnagel: +1 - this would be a breaking change for Core if I'm not mistaken (not just identity-config) because the restriction would apply on an upgrade (being part of the plugin code). This _might_ meet the bar of a security relevant change that will not require a major version bump (per our [versioning policy](https://docs.defenseunicorns.com/core/reference/policies/versioning/#security-exception)) but definitely still want to be careful about how we introduce and announce this.

mjnagel: (for example) as an alternative, we could enforce this hardening rule via Pepr validations, which is an _expansion_ of the current posture (allowing new types of public clients beyond device flow). This wouldn't impact someone's ability to make clients a different way separate from the operator, and thus wouldn't be a breaking change, but would still provide a degree of hardening (no different than what we've done in the past).

slaskawi: I do agree with you that this might be a breaking change as the UDS Client Policy is also executed if a System Administrator creates a new Client via Tofu or ClickOps.

The validation performed here is rather a set of good OAuth Client configuration practices and I think we could approach it more generally - as @mjnagel suggested - in Pepr. Implementing this as a Policy (with the Exemption mechanism) would be definitely more flexible. 

I've created [CORE-498](https://linear.app/defense-unicorns/issue/CORE-498/uds-operator-policy-for-validating-keycloak-client) to track this effort and in the meantime, let me remove this additional validation logic from here.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
slaskawi: You're right - this is definitely a mistake. 

But either way - I'm removing this piece for now (see the previous comment).

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
mjnagel: I might be missing something, but I think in the current state this is still a breaking change?

Previously: 
- Package CR validation: only device flow public clients allowed
- Keycloak: Public clients allowed with no restrictions

Now:
- Package CR validation: device flow + pkce public clients allowed (expanded)
- Keycloak: Only device flow allowed by default, pkce allowed if enabled

So on an upgrade of an existing Keycloak users would no longer have non-device flow clients allowed until they manually enable the toggle for this. Even after enabling the toggle they wouldn't be able to create public clients without PKCE (which they could've before). I think from a security standpoint enforcing PKCE is _probably fine_ to do even in a minor (although mentioned as a breaking change). I struggle more with this defaulting to disabled on existing installs (from a breaking change perspective).

joelmccoy: I have the same sentiment here. I only imagined validation changes on the Pepr side to support this feature for operator created clients.  Enforcing this policy for all clients would definitely be a breaking change and I'm not sure was the scope of the original ask? I'd vote for keeping it simple and doing all the validations on the Operator side. Happy to sync and chat this through a bit too.  I haven't done a huge deep dive on this code and would to get the approach agreed upon first.

slaskawi: Agree and I fully refactored the code. Should be better now.

mjnagel (commented):
LGTM pending the merge on core side.

mjnagel (approved):

PR #870, opened 2026-05-29
fix(ci): e2e tests breaking because of envoyfilter hanging on grafana

joelmccoy:
## Description

Our UDS Core envoy filter for classification banner is breaking on Grafana.  We set this feature on for the slim-dev bundle but we don't test it in this repo.  Removing this from here since we should test this UDS Core instead.

Create issue to address this in core and test there instead: https://linear.app/defense-unicorns/issue/CORE-590/classification-banner-envoyfilter-hangs-on-streamedspa-responses-eg

Additionally removed the fips field from the bundle as that no longer is a supported value.

## Type of change

## Checklist before merging


chance-coleman (approved):

slaskawi (commented):
LGTM!

PR #871, opened 2026-06-02
feat: add usercertificateSKI user attribute from x509 cert

JeffResc:
## Description
Extracts the SKI (X.509 extension OID `2.5.29.14`) from the presented client cert during Keycloak registration and the x509 update flow, and stores it on the user as a new `usercertificateSKI` attribute, mirroring how `usercertificateCN` works today.

Cert-derived attributes (both `usercertificateCN` and `usercertificateSKI`) are now only persisted when the presented x509 passes validation (`x509Username != null`)

## Type of change

## Checklist before merging


src/realm.json5
```diff
@@ -2136,7 +2136,7 @@
...
...nization or Company Name\",\"validations\":{},\"annotations\":{},\"permissions\":{\"view\":[\"admin\",\"user\"],\"edit\":[\"admin\",\"user\"]},\"multivalued\":false},{\"name\":\"location\",\"displayName\":\"Anti-bot Location (42)\",\"validations\":{},\"annotations\":{},\"permissions\":{\"view\":[\"admin\"],\"edit\":[\"admin\",\"user\"]},\"multivalued\":false},{\"name\":\"notes\",\"displayName\":\"Access Request Notes\",\"validations\":{},\"annotations\":{\"inputType\":\"textarea\"},\"permissions\":{\"view\":[\"admin\",\"user\"],\"edit\":[\"admin\",\"user\"]},\"multivalued\":false},{\"name\":\"usercertificateIdentity\",\"displayName\":\"x509 User Identity\",\"validations\":{},\"annotations\":{},\"permissions\":{\"view\":[\"admin\"],\"edit\":[\"admin\"]},\"multivalued\":false},{\"name\":\"usercertificateCN\",\"displayName\":\"x509 User CN\",\"validations\":{},\"annotations\":{},\"required\":{\"roles\":[\"user\"]},\"permissions\":{\"view\":[\"admin\"],\"edit\":[\"admin\"]},\"multivalued\":false}],\"groups\":[{\"name\":\"user-metadata\",\"displayHeader\":\"User metadata\",\"displayDescription\":\"Attributes, which refer to user metadata\"}],\"unmanagedAttributePolicy\":\"ADMIN_EDIT\"}"
```
chance-coleman, line 2139: nice, admin only view/edit is definitely the right call here.

src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/X509ToolsTest.java
```diff
@@ -114,6 +114,27 @@ public void testGetX509CommonNameFromFormContextNull() throws Exception {
         assertNull(commonName);
     }
 
+    @Test
```
chance-coleman, line 117: Tests covering happy path, which is good. I'd probably advocate for an additional test for an absent SKI extension or malformed extension.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/Common.java
```diff
@@ -71,6 +71,17 @@ public interface Common {
      */
     String USER_X509_CN_ATTRIBUTE = "usercertificateCN";
 
+    /**
+     * The user id attribute tracks the Keycloak attribute that maps to the user's
```
chance-coleman: is `user id attribute` accurate here? i think its kind of convoluted with other attributes. I'd suggest updating the comment to something like:

```suggestion
     * The user certificate SKI attribute tracks the Keycloak attribute that maps to the user's
```

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/RegistrationValidation.java
```diff
@@ -54,6 +54,10 @@ private static void processX509UserAttribute(final RealmModel realm, final UserM
             // Bind the X509 Common name attribute to the user
             user.setSingleAttribute(Common.USER_X509_CN_ATTRIBUTE, x509commonName);
         }
+        if (x509subjectKeyId != null) {
```
chance-coleman, line 57: If I’m reading this correctly, `usercertificateSKI` can be written even when `x509Username` is `null`. `getX509SubjectKeyId()` extracts from any presented cert, but it does not go through the existing cert-policy/user-identity validation path.

If anything downstream treats `usercertificateSKI` as a trusted CAC binding, an invalid or unaccepted cert could still populate it. I think we should only extract/write SKI when `x509Username != null`, or gate both CN and SKI behind the same accepted-X509 check.

chance-coleman (commented):
Direction is looking good! Couple smaller comments that i think would be good to address.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/Common.java
bburky: I might actually spell out Subject Key Identifier. "usercertificateSKI" does't mean anything and I had to google by OID number to figure out what this was

```suggestion
     * The user Subject Key Identifier attribute tracks the Keycloak attribute that maps to the user's
```

I think it's fine to keep it short in the actual field name, but spell it out in the UI and longer comments probably

slaskawi:
@codex review

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/X509Tools.java
```diff
@@ -316,6 +319,48 @@ public static String getX509CommonName(final KeycloakSession session, final Http
         return null;
     }
 
+    /**
+     * Extracts the Subject Key Identifier (SKI) from the x509 certificate.
+     */
+    public static String getX509SubjectKeyId(final FormContext context) {
```
slaskawi: It looks like other methods in this class usually take the `KeycloakSession session` and `HttpRequest httpRequest` pair as an argument. Could we do the same here for the symmetry reasons? I think it will also spare a few lines of code as a side effect :)

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/RegistrationValidation.java
JeffResc: Good catch, the implementation has been updated so that `usercertificateCN` and `usercertificateSKI ` are only persisted when `x509Username != null`

src/plugin/src/test/java/com/defenseunicorns/uds/keycloak/plugin/X509ToolsTest.java
JeffResc: I added `testGetX509SubjectKeyIdAbsentExtension` and `testGetX509SubjectKeyIdMalformedExtension`, both returning `null`

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/X509Tools.java
JeffResc: I removed the `FormContext/RequiredActionContext`overloads, `getX509SubjectKeyId` now takes `(KeycloakSession, HttpRequest)` args

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/RegistrationValidation.java
```diff
@@ -62,8 +66,9 @@ public void success(final FormContext context) {
         RealmModel realm = context.getRealm();
         String x509Username = X509Tools.getX509Username(context);
         String x509commonName = X509Tools.getX509CommonName(context);
+        String x509subjectKeyId = X509Tools.getX509SubjectKeyId(context);
```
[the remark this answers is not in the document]
chance-coleman: This is addressed by my earlier comment.

chance-coleman (approved):
This looks good from my perspective. @slaskawi 's approval would also be appreciated here. Then we can get this in the next release of uds-core (Tuesday the 9th).

slaskawi (approved):

PR #877, opened 2026-06-10
feat: add uds-fleet-admin client and implement jwt auth with issuer auto-discovery

joelmccoy:
## Description

Adds the `uds-fleet-admin` Keycloak client for use with Fleet Command.  This client used Keycloak Kubernetes JWT Authentication.  This client is declared in the realm init and for already running environments there will be manual upgrade steps.

> [!NOTE]
> This PR adds a custom plugin to do issuer auto-discovery and determines whether or not to send the auth header to the jwks endpoint.  This logic is going to be added upstream in https://github.com/keycloak/keycloak/pull/50224 and the plugin logic here should generally match that PR. We need to write a custom plugin for this in the meantime as we don't expect this logic to land in Keycloak for a couple months and we need this feature now.  Once that feature lands we can completely delete this plugin and just rely on the upstream `kubernetes` identity provider.

* Adds custom `uds-kubernetes` IDP that extends upstream `Kubernetes` for JWT Auth in realm config
* Adds Federated JWT Auth as an accepted Client Authentication method in realm config
* Adds `uds-fleet-admin` client with proper permissions
* Updates UDS Client Policy Plugin to scope `uds-fleet-admin` to only be allowed to create/update/delete clients with the `fleet-*` prefix (note: Technically this client can do a wholesale list of all client metadata see [here](https://github.com/defenseunicorns/uds-identity-config/pull/877#discussion_r3391757628), but this is an acceptable risk)
* Adds e2e tests for managing fleet clients with this Client via JWT flow
* Adds auto-discovery logic for issuer (this is needed for managed EKS clusters who don't use the in cluster token issuer)
* Updates jwks fetch requests so that it does not send a k8s authorization header to jwks endpoints out of the cluster

I tested this on EKS/RKE2/AKS on this pr as well: https://github.com/defenseunicorns/uds-core/pull/2734.

Follow up docs PR for manual upgrade steps will be done in a follow up UDS Core PR.

## Related Issue

Relates to CORE-598

## Type of change

## Checklist before merging


src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -42,20 +42,29 @@
...
+    public static final String ATTRIBUTE_CREATED_BY = "created-by";
+
+    /**
+     * Clients governed by this executor, mapped to the clientId prefix required for clients they create.
+     * An empty prefix imposes no naming constraint (preserves uds-operator's existing behavior).
+     */
+    public static final Map<String, String> MANAGED_CLIENT_ID_PREFIXES = Map.of(
+            "uds-operator", "",
+            "uds-fleet-admin", "fleet-"
+    );
```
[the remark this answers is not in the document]
joelmccoy: @greptileai this is a good callout, but I'm pretty sure the listing permissions do not include secrets, just metadata on the client.  There is no way with our current plugin to block this enumeration.  I do think this is an ok and acceptable risk.

src/test/cypress/support/commands.ts
```diff
@@ -239,15 +249,54 @@ Cypress.Commands.add("getAccessToken", (subject?: TokenSubject) => {
...
+    expect(wellKnown.status).to.eq(200);
+    const issuer = wellKnown.body.issuer as string;
+    expect(issuer, "realm issuer").to.be.a("string").and.not.be.empty;
+
+    // Mint a short-lived projected SA token addressed to the realm issuer.
+    const mintCmd = `uds zarf tools kubectl create token ${saName} -n ${saNamespace} --audience "${issuer}" --duration=10m`;
+    return cy.exec(mintCmd).then((result) => {
+      expect(result.exitCode, "kubectl create token").to.eq(0);
+      const saToken = result.stdout.trim();
+      expect(saToken, "service account token").to.be.a("string").and.not.be.empty;
```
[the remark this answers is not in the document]
joelmccoy: suppressed this log

src/test/cypress/support/commands.ts
```diff
@@ -180,6 +180,16 @@ Cypress.Commands.add("getClientSecret", (clientId: string) => {
...
 
+/**
+ * Assert a JWT access token is a string carrying the manage-clients role, then return it.
+ */
+function assertManageClientsToken(accessToken: string): string {
+  expect(accessToken).to.be.a("string");
+  const tokenPayload = Buffer.from(accessToken.split(".")[1], "base64").toString("utf-8");
+  expect(tokenPayload).contains("manage-clients");
+  return accessToken;
+}
```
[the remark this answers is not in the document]
joelmccoy: added a helper to do this conversion

src/test/cypress/support/commands.ts
```diff
@@ -239,15 +249,54 @@ Cypress.Commands.add("getAccessToken", (subject?: TokenSubject) => {
...
+ *
+ * Mirrors how a real Fleet pod authenticates: a Kubernetes service-account token,
+ * addressed (audience) to the realm issuer, is presented as a client assertion.
+ * Keycloak matches it to the uds-fleet-admin client by the configured IdP issuer + SA "sub".
+ */
+Cypress.Commands.add("getFleetAdminAccessToken", () => {
+  const realmTokenUrl = "https://keycloak.admin.uds.dev/realms/uds/protocol/openid-connect/token";
+  const wellKnownUrl = "https://keycloak.admin.uds.dev/realms/uds/.well-known/openid-configuration";
+  const saNamespace = "uds-fleet-command";
+  const saName = "uds-fleet-command-sa";
```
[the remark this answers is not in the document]
joelmccoy: ehh overkill and not the pattern we follow for other tests. not taking this suggestion.

src/realm.json5
```diff
@@ -909,6 +930,36 @@
             "authenticationFlowBindingOverrides": {},
             "fullScopeAllowed": true,
             "nodeReRegistrationTimeout": 0
+        },
+        {
+            "id": "94de6a0d-83f7-4875-93f0-05a986ecc3a5",
+            "clientId": "uds-fleet-admin",
+            "name": "uds-fleet-admin",
+            "description": "Scoped admin client for UDS Fleet Management to manage fleet-* clients",
+            "enabled": "${REALM_FLEET_CLIENT_ENABLED:false}",
```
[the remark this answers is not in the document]
joelmccoy: REALM is prepended to realInitEnv values. This is correct.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -42,20 +42,29 @@
...
- * The following changes are introduced to the Client:
- * * The Client is marked as created by the UDS Operator by setting the attribute "created-by=uds-operator".
- * * The UDS Operator's Token is checked if it can access particular Client (the client need to contain "created-by=uds-operator" attribute).
+ * The executor only acts when the request is authenticated as one of the managed clients
+ * ({@link #MANAGED_CLIENT_ID_PREFIXES}). For those clients the following rules apply:
+ * * Clients created by a managed client are stamped with "created-by=&lt;authenticated clientId&gt;".
+ * * A managed client may only update/view/delete clients it created (matching "created-by" attribute).
+ * * A managed client may only create clients whose clientId starts with its required prefix (empty = no constraint).
  * * The Client can't use the Full Scope Allowed feature.
  * * The Client can only use the Protocol Mappers that are whitelisted in the configuration. The default, opinionated configuration uses OIDC nad UDS Protocol Mappers.
```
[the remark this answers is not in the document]
joelmccoy: fixed.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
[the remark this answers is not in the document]
joelmccoy: Leaving this open for future reviewers to see. I think this approach is fine.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -42,20 +42,29 @@
...
-    public static final String ATTRIBUTE_UDS_OPERATOR = "created-by";
-    public static final String ATTRIBUTE_UDS_OPERATOR_VALUE = "uds-operator";
+    public static final String ATTRIBUTE_CREATED_BY = "created-by";
+
+    /**
+     * Clients governed by this executor, mapped to the clientId prefix required for clients they create.
+     * An empty prefix imposes no naming constraint (preserves uds-operator's existing behavior).
+     */
+    public static final Map<String, String> MANAGED_CLIENT_ID_PREFIXES = Map.of(
+            "uds-operator", "",
```
slaskawi, line 67: Do we need a map here? It seems the mechanism could be simplified (as implied in the Technical Design) to the authenticated client with no additional remapping. 

This attribute is not viewable in the Admin UI, so we can put whatever we want here.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
```diff
@@ -108,31 +117,30 @@ private static String toString(ClientRepresentation rep) {
     @Override
     public void executeOnEvent(ClientPolicyContext context) throws ClientPolicyException {
         if ((context instanceof ClientCRUDContext clientCRUDContext)) {
-            logger.debug("Executing UDSClientPolicyPermissionsExecutor, authenticated to Client ID: {}, ", getAuthenticatedClientId(clientCRUDContext));
+            String authenticatedClientId = getAuthenticatedClientId(clientCRUDContext);
+            logger.debug("Executing UDSClientPolicyPermissionsExecutor, authenticated to Client ID: {}", authenticatedClientId);
 
-            if (clientCRUDContext.getAuthenticatedClient() != null && UDS_OPERATOR_CLIENT_ID.equals(getAuthenticatedClientId(clientCRUDContext))) {
+            if (clientCRUDContext.getAuthenticatedClient() != null && authenticatedClientId != null && MANAGED_CLIENT_ID_PREFIXES.containsKey(authenticatedClientId)) {
```
slaskawi: Could we add all these sub-conditions to the `logger.debug` line the above? This will signifficantly help with debugging potential issues in the future.

src/test/cypress/support/commands.ts
```diff
@@ -239,15 +257,55 @@ Cypress.Commands.add("getAccessToken", (subject?: TokenSubject) => {
...
+
+  // The SA token must be addressed to the realm issuer (audience). Read it from the realm's discovery doc
+  // so the audience always matches what Keycloak advertises, regardless of the configured frontend URL.
+  return cy.request({ method: "GET", url: wellKnownUrl }).then((wellKnown) => {
+    expect(wellKnown.status).to.eq(200);
+    const issuer = wellKnown.body.issuer as string;
+    expect(issuer, "realm issuer").to.be.a("string").and.not.be.empty;
+
+    // Mint a short-lived projected SA token addressed to the realm issuer.
+    const mintCmd = `uds zarf tools kubectl create token ${saName} -n ${saNamespace} --audience "${issuer}" --duration=10m`;
```
slaskawi: nit: Duration can be skipped - it's 10m by default.

src/realm.json5
```diff
@@ -521,6 +521,17 @@
...
             }
+        },
+        {
+            "id": "00155696-4be6-4807-8df7-b5650555e0ea",
+            "username": "service-account-uds-fleet-admin",
+            "enabled": true,
+            "serviceAccountClientId": "uds-fleet-admin",
+            "clientRoles": {
+                "realm-management": [
+                    "manage-clients"
```
slaskawi, line 532: Just wanted to double check - are we going to need Realm Roles? The UDS Fleet will create potentially 10s of Clients - they might want to share some roles across them. 

Since confirming might take a while, I'm OK to proceed without this information but please double check it in-between.

src/realm.json5
```diff
@@ -540,6 +551,16 @@
                     "view-groups"
                 ]
             }
+        ],
+        // uds-fleet-admin runs with fullScopeAllowed=false, so its service-account roles are filtered
+        // to the client's scope. Grant only manage-clients into scope so it reaches the access token.
+        "realm-management": [
```
slaskawi: This looks suspicious! Why do we need it exactly? The UDS Operator uses similar configuration and it doesn't need this. Why?

slaskawi (commented):
Fantastic work @joelmccoy ! Added a few comments and once addressed, I'm happty to approve.

src/realm.json5
joelmccoy: I reviewed the current Fleet client provisioning. It only creates OIDC clients for each agent, retrieves their generated client secrets, and then deletes those clients. Currently, the payload does not create or assign realm roles or client roles and these fleet clients shouldn't need to create roles to manage things in keycloak.

src/realm.json5
joelmccoy: this is because i set `"fullScopeAllowed": false` on the client and you need to be explicit with which roles are granted for the client's scope.

However, I now realize this approach is not really necessary. I am going to just mirror the operator client and set `"fullScopeAllowed": true` so we don't need this.  Setting this to true will just grant whatever roles the client has assigned which is just `manage-clients`

src/realm.json5
```diff
@@ -909,6 +920,36 @@
             "authenticationFlowBindingOverrides": {},
             "fullScopeAllowed": true,
             "nodeReRegistrationTimeout": 0
+        },
+        {
+            "id": "94de6a0d-83f7-4875-93f0-05a986ecc3a5",
+            "clientId": "uds-fleet-admin",
+            "name": "uds-fleet-admin",
+            "description": "Scoped admin client for UDS Fleet Management to manage fleet-* clients",
+            "enabled": "${REALM_FLEET_CLIENT_ENABLED:false}",
```
joelmccoy, line 929: Does anyone have any oppositions to setting this default to true instead?  Would be better UX for fleet folks as the won't need to provide an manual override in their bundles?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
joelmccoy: We don't need a map for the created-by value; it should stay as the authenticated client ID. However, we still require a configuration structure to identify which authenticated clients are managed by this executor and the prefix rule that applies to each.

src/test/cypress/support/commands.ts
joelmccoy: removed

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/clientpolicy/executor/UDSClientPolicyPermissionsExecutor.java
joelmccoy: added some debug logging here

joelmccoy:
@greptileai review

src/realm.json5
joelmccoy: going to keep this off by default given internal discussion.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProviderConfig.java
```diff
@@ -0,0 +1,78 @@
...
+        return !Boolean.FALSE.toString().equals(getConfig().get(AUTOMATIC_ISSUER_DISCOVERY));
+    }
+
+    /**
+     * Validation hook (fires on IdP create/update, including realm import). Resolves and persists the issuer
+     * before the standard issuer-uniqueness validation.
+     */
+    @Override
+    public void validate(RealmModel realm) {
+        resolveAndPersistIssuer();
```
chance-coleman, line 60: If the API is unreachable at realm import time, this throws and the import fails entirely. Would it make sense to have a retry here to avoid a start race conditiom?

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProviderConfig.java
```diff
@@ -0,0 +1,78 @@
...
+        super.validate(realm);
```
chance-coleman, line 61: I don't think it actually makes much difference, so mostly for awareness, the `validate()`  and `resolveAndPersistIssuer()` are flipped in order in the upstream PR. This is so Keycloak's own SSL-required and URL format checks run before any outbound HTTP

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/KubernetesUtils.java
```diff
@@ -0,0 +1,203 @@
...
+            throw e;
+        } catch (Exception e) {
+            throw new IllegalArgumentException("Failed to resolve Kubernetes issuer from " + wellKnown, e);
+        }
+    }
+
+    /**
+     * Read the mounted Keycloak pod service-account token, or null if it is not mounted or unreadable.
+     */
+    static String readServiceAccountToken() {
```
chance-coleman, line 179: Upstream adds `SERVICE_ACCOUNT_TOKEN_PATH_PROPERTY` so tests can redirect to a temp file.  When this plugin is deleted and replaced by the stock provider, this test pattern breaks entirely. A two-line backport `(System.getProperty(SERVICE_ACCOUNT_TOKEN_PATH_PROPERTY, SERVICE_ACCOUNT_TOKEN_PATH))` would close the gap with upstream and remove the static-mock dependency.

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/KubernetesUtils.java
```diff
@@ -0,0 +1,203 @@
...
+            return false;
+        }
+        return "https".equals(jwksUri.getScheme())
+                && "/openid/v1/jwks".equals(jwksUri.getPath())
+                && jwksUri.getQuery() == null
+                && jwksUri.getFragment() == null
+                && isIpLiteral(jwksUri.getHost());
+    }
+
+    private static boolean isIpLiteral(String host) {
```
chance-coleman, line 130: Any string with a colon goes to `getByName()`, which can do a DNS lookup. I'm concerned about the security boundary here since a specially crafted JWKS URI only needs to contain a `:` to get by this.

chance-coleman (commented):
Generally looking good, few small questions and comments 🚀

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProviderConfig.java
joelmccoy: is this the same for upstream though? also won't the pod just crash and just recover hopefully?  either way, i'd want to mirror with the approach taking upstream

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/UDSKubernetesIdentityProviderConfig.java
joelmccoy: Upstream can change the order because it completely rewrote `validate()`. We delegate to the stock `super.validate()`, which validates the issuer. Therefore, we need to resolve and persist before calling it. Changing the order would require copying the entire upstream method body just to reorder, which isn't worth it for a temporary plugin I think

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/KubernetesUtils.java
joelmccoy: fix in: https://github.com/defenseunicorns/uds-identity-config/pull/877/commits/4c98e5ebe40d04f509ddc357548f8523c562e858 with a library. you can probably port this upstream too in your PR

src/plugin/src/main/java/com/defenseunicorns/uds/keycloak/plugin/broker/kubernetes/KubernetesUtils.java
joelmccoy: adjusted to match upstream

chance-coleman (approved):
LGTM, happy with where this is at!

